PR #2, opened 2022-08-01
MLPAB-36: Add dockerised Go app

acsauk:
Adding some initial stubs so we can play a couple more setup tickets that rely on an app/Go code being in place for them to run (testing, pre-commit hooks etc).


Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM golang:1.18 as build-env
+
+WORKDIR /app
+
+COPY go.mod .
+
+RUN go mod download
+
+COPY . .
```
andrewpearce-digital, line 9: it would be good to put all this starter for an image in a subdirectory, otherwise when this and PR#1 are merged this copy is going to grab everything in the repo

README.md
```diff
@@ -1,2 +1,9 @@
 # opg-modernising-lpa
-OPG Modernising LPA: Managed by opg-org-infra &amp; Terraform
+## Getting Started
+### Prerequisites
+* Docker and docker-compose
+
+### Installation
+```bash
+$ docker-compose up -d
```
andrewpearce-digital, line 8: docker now supports a copmpose command. It's suprisingly easier to use than finding - :D 


```suggestion
$ docker compose up -d
```

andrewpearce-digital (approved):

hawx (approved):

andrewpearce-digital (approved):

PR #15, opened 2022-08-04
MLPAB-38: Build assets and configure GDS/MOJ Frontend

acsauk:
# Purpose

Adds a dummy page to prove assets are compiled, served, and accessible from our Go app. Includes boilerplate SCSS and JS to enable GDS and MOJ frontend frameworks.

Fixes MLPAB-38

Screenshot 2022-08-04 at 12 31 43

## Approach
I had to do quite a few steps here at once in order to prove everything was hooked up and configured correctly. As I'm getting my head back into Go I decided against using some of the `opg-common` packages we have available so I could be sure I understood how certain aspects were working. We can move to using the common packages when we start developing proper.

This change includes:
*  Adding an assets build stage to Dockerfile
* Init-ing JS using yarn
* Installing GDS and MOJ frontend packages
* Serving a dummy page on localhost:5050/home to show the frontends are initialised correctly
* Serving static assets using a `http.FileServer` (this is open to discussion - just wanted a quick way to make things available)

## Learning

Dipped into Alex Edwards [Let's Go!](https://lets-go.alexedwards.net/) for pointers on serving static assets and to remind me about route handlers in Go.

## Checklist


app/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "build:images": "mkdir -p web/static/assets/images && cp node_modules/govuk-frontend/govuk/assets/images/* node_modules/@ministryofjustice/frontend/moj/assets/images/* web/static/assets/images",
+    "build:fonts": "mkdir -p web/static/assets/fonts && cp node_modules/govuk-frontend/govuk/assets/fonts/* web/static/assets/fonts",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "@ministryofjustice/frontend": "^1.4.2",
+    "dart-sass": "^1.25.0",
+    "govuk-frontend": "^4.2.0",
+    "jquery": "^3.6.0"
```
johnnolan, line 17: We could probably drop the jquery dependency and make the build even simpler if we just got the compiled JS from GDS like here https://github.com/mattmachell/esbuild-boilerplate/blob/main/example/main.js#L1 what do you think?

web/template/layout/base.gohtml
```diff
@@ -0,0 +1,169 @@
...
+                    </div>
+                </div>
+            </div>
+        </footer>
+
+        <script src="{{ .Prefix }}/static/javascript/main.js"></script>
+    </body>
+
+</html>
+{{ end }}
```
johnnolan, line 169: A few missing end of file lines in the files committed

app/main.go
```diff
@@ -2,14 +2,62 @@ package main
...
+	"net/http"
 )
 
 func Hello() string {
 	return "Hello, world!"
 }
 
+type PageData struct {
+	WebDir      string
+	Prefix      string
```
hawx, line 17: since this will be always be hosted at the root (right?) we shouldn't need a prefix (and I've kind of hated how it has spread everywhere)

app/main.go
```diff
@@ -2,14 +2,62 @@ package main
 
 import (
 	"fmt"
-	"github.com/ministryofjustice/opg-modernising-lpa/internal/fake"
+	"github.com/ministryofjustice/opg-go-common/env"
+	"html/template"
+	"log"
+	"net/http"
```
hawx, line 8: if you set up https://pkg.go.dev/golang.org/x/tools/cmd/goimports instead of `gofmt` it will nicely sort and break up the imports

hawx (approved):
obviously feel free to ignore my comments if you just want to get something up and running, it can all be changed later

app/package.json
acsauk: I shuddered at having to install it but if we want to rely on any components in the MOJ design system it looks like we need it - https://design-patterns.service.justice.gov.uk/get-started/setting-up-javascript/ (unless there's a similar workaround for MOJDS as well?)

app/main.go
```diff
@@ -2,14 +2,62 @@ package main
...
+	prefix := env.Get("PREFIX", "")
+
+	data := PageData{
+		WebDir:      webDir,
+		Prefix:      prefix,
+		ServiceName: "Modernising LPA",
+	}
+
+	files := []string{
+		fmt.Sprintf("%s/template/home.gohtml", webDir),
```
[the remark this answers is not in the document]
acsauk: Ooo nice not used that before - will switch out 👍

app/main.go
```diff
@@ -2,14 +2,62 @@ package main
...
+		fmt.Sprintf("%s/template/layout/base.gohtml", webDir),
+	}
+
+	ts, err := template.ParseFiles(files...)
```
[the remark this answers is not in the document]
acsauk: Agree - I'd gone with hand rolling it, for now, to just get the hang of how templating works as I still had a Symfony/PHP approach when I started this ticket of just putting templates in a folder and expecting Go to know where they were 😂 I think once we start on the first template we can drop this and bring in the common package to set it up.

app/main.go
acsauk: Cool - I was struggling to see how this worked in `opg-sirius-lpa-frontend` where I pinched it from so I'm more than happy to drop it out.

app/main.go
acsauk: I had this in the pre-commit hook setup but then realised on my last commit it hadn't been running. Looks like its working now though (had to re-install it) so I'll fix this up 👍

app/package.json
johnnolan: Leave it in for now then and we can always revisit it as we start to build more of the journey etc, all looking good mate!

johnnolan (approved):

PR #17, opened 2022-08-05
MLAPB-22: Add cypress and codecept

acsauk:
# Purpose

Adds cypress and codeceptJS as potential e2e/integration test frameworks and initial tests to cover the dummy page we have right now.

Fixes MLPAB-22

## Approach

I've included codeceptJS here as an alternative to cypress so we can evaluate over a couple of tickets and then decide which we want to go with long-term. At first glance, I prefer the API of codeceptJS to cypress, but until we apply it to real code it's too early to say if this trumps the more mature ecosystem around cypress.

I've included a way to run accessibility checks in both frameworks. Cyrpess was super simple as they have an official plugin but for codeceptJS I've had to lean on a playwright plugin which ties us down to a single driver driving our tests. There may well be a way to generalise that step but the maintenance cost is something to bear in mind when we come to evaluate both frameworks.

I moved the JS files out of the `app` folder as it keeps a clear separation between Go and JS code.

## Learning

Both testing frameworks have decent docs and SO helped with discovering the playwright-axe package.

## Checklist


app/package.json
```diff
@@ -1,23 +1,41 @@
...
+        "cypress:open": "cypress open",
+        "cypress:run": "cypress run",
+        "cypress:run:debug": "cypress run --headed",
+        "test": "echo \"Error: no test specified\" && exit 1",
+        "codecept": "codeceptjs run --steps",
+        "codecept:headless": "HEADLESS=true codeceptjs run --steps",
+        "codecept:ui": "codecept-ui --app",
+        "codecept:demo": "codeceptjs run --steps -c node_modules/@codeceptjs/examples",
+        "codecept:demo:headless": "HEADLESS=true codeceptjs run --steps -c node_modules/@codeceptjs/examples",
+        "codecept:demo:ui": "codecept-ui --app  -c node_modules/@codeceptjs/examples"
```
acsauk, line 19: Leaving these here for the time being so others can see what more involved tests look like

app/cypress/e2e/confirm_components_rendered.cy.js
```diff
@@ -0,0 +1,29 @@
...
+
+    it('displays a GDS summary element', () => {
+        cy.checkA11y()
+
+        cy.get('summary').first()
+            .should('contain.text', 'Help with nationality')
+    })
+
+    it('displays a MOJ password reveal element', () => {
+        cy.checkA11y()
```
mattmachell, line 15: Be tempted to move this into an after hook or before hook.

app/codecept/scenarios/gds_moj_components_render_test.js
```diff
@@ -0,0 +1,18 @@
...
+Scenario('displays a GDS summary element', ({ I}) => {
+    I.amOnPage('/home')
+    I.seeElement(
+        locate('summary').withText('Help with nationality')
+    )
+    I.runAccessibilityChecks()
+});
+
+Scenario('displays a MOJ password reveal element', ({ I}) => {
+    I.amOnPage('/home')
```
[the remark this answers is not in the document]
acsauk: Yeah `I` is the [base actor object](https://codecept.io/basics/#getting-started) that all of the steps branch off in codecept. We can try to give it context when/if we have the concept of user roles by extending in custom commands but the API may look a bit weird as its designed to be used in a first-person perspective (e.g. `Admin.amOnPage()`).

app/cypress/e2e/confirm_components_rendered.cy.js
acsauk: Makes sense 👍 

Once we have a few tests in place I'd be interested in seeing the performance hit of running this as may be better to selectively run it if it's slowing run times down.

app/codecept/scenarios/gds_moj_components_render_test.js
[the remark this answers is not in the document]
mattmachell: Yeah, one of the reasons I dislike this API structure is that it's brittle when you don't write tests exactly as the authors expect. I suppose we can do things like `I.logInAsAdmin()` to adapt to it, but there'll likely be other issues too.

app/codecept/scenarios/gds_moj_components_render_test.js
```diff
@@ -0,0 +1,18 @@
+Feature('GDS and MOJ components are available');
+
+Scenario('displays a GDS summary element', ({ I}) => {
+    I.amOnPage('/home')
+    I.seeElement(
+        locate('summary').withText('Help with nationality')
```
mattmachell, line 6: Locating via text is a pretty brittle pattern in e2e tests. Might be worth getting test hooks via data attributes and a constants file in place as part of this work to avoid issues later.

app/package.json
```diff
@@ -1,23 +1,41 @@
...
+    },
+    "license": "MIT",
+    "dependencies": {
+        "@ministryofjustice/frontend": "^1.4.2",
+        "dart-sass": "^1.25.0",
```
mattmachell, line 24: `dart-sass` and `sass` should be the same library if you use an up-to-date sass version. It should probably only be in dev deps too unless you're doing runtime compiling.

app/codecept/scenarios/gds_moj_components_render_test.js
acsauk: In this example, I'd agree but I think there are some situations where locating by text is actually helpful as its giving an assertion for free on if something has gone wrong with rendering text on a component that has a design that is expected to be static.

As much as I don't like adding things to the DOM just for testing purposes I can see the value in some high-level data-* attributes we can hook into where we need to. Do you have some examples of how other teams have approached this?

app/package.json
acsauk: Ah good point - I was on autopilot following the design systems install steps. Will try dropping out and confirm everything builds OK.

app/codecept/scenarios/gds_moj_components_render_test.js
mattmachell: It comes down a bit to "what are you testing with this?" Sometimes you might want to check the text is "right", so on translations and errors, but other times it's outside the scope. Also means message catalogue changes are coupled to deploys, which you might not want if you have a bigger content/translation team working outside of the dev process.

I put some examples of data attribute use in the testing with cypress doc. They don't need to be test specific, in fact I avoid `data-qa-hook` personally, I tend toward the pattern of additional app semantics, like `data-action="client.1.edit"` which convey intent of the page element.

acsauk: I'll leave this test as is as it's here to prove everything is hooked up right and will be dropped when we write our first real test but will apply the `data-*` approach for the login flow.

acsauk: I'll leave this test as is as it's here to prove everything is hooked up right and will be dropped when we write our first real test but will apply the data-* approach for the login flow.

hawx (approved):

PR #25, opened 2022-08-12
MLAPB-39: Add Gov UK Sign In

acsauk:
# Purpose

Adds the integration code for signing in via Gov UK Sign In along with a mocked implementation of Gov UK Sign In to use locally and for testing.

Fixes MLPAB-39

## Approach

As we don't have a data store set up yet this implementation goes through the OIDC flow and parses the user's email to a page. There are also a few more steps we will want to do before this is fully working:

* Cache JWT/access tokens
* Use refresh tokens to refresh access once the original token has expired
* Validate standard claims before decoding/using a JWT (sub, iss, nonce, aud, exp, iat and alg header) 

I also dropped Codecept as part of this PR. There were two main reasons for doing this (or three if you include @hawx preferring cypress 😂) :

* The underlying dependencies are already being flagged by dependabot as having vulnerabilities by relying on some quite old versions. This doesn't bode well for the level of maintenance we would have to carry out.
* I couldn't find a way of supporting multi-domain tests whereas cypress supports this [out the box](https://www.cypress.io/blog/2022/04/25/cypress-9-6-0-easily-test-multi-domain-workflows-with-cy-origin/). We don't rely on it in this PR but if we ever wanted to write a smoke test against the Gov UK Sign In integration env we would need to use another framework.

## Checklist


app/internal/signin/get_token.go
```diff
@@ -0,0 +1,119 @@
...
+type TokenResponseBody struct {
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+	TokenType    string `json:"token_type"`
+	ExpiresIn    int    `json:"expires_in"`
+	IdToken      string `json:"id_token"`
+}
+
+func (c *Client) GetToken(redirectUri, clientID, JTI, code string) (string, error) {
+	log.Println("GetToken()")
```
hawx, line 38: won't need these logs

app/internal/signin/get_token.go
```diff
@@ -0,0 +1,119 @@
...
+
+	body := &TokenRequestBody{
+		GrantType:           "authorization_code",
+		AuthorizationCode:   code,
+		RedirectUri:         redirectUri,
+		ClientAssertionType: "urn:ietf:params:oauth:client-assertion-type:jwt-bearer",
+		ClientAssertion:     signedAssertion,
+	}
+
+	encodedPostBody := new(bytes.Buffer)
```
hawx, line 64: `var encodedPostBody bytes.Buffer` is the usual way of doing this

app/main.go
```diff
@@ -38,13 +68,27 @@ func main() {
...
+	err = signInClient.Discover(issuer.String() + "/.well-known/openid-configuration")
+	if err != nil {
+		logger.Fatal(err)
+	}
 
 	mux.Handle("/static/", http.StripPrefix("/static", fileServer))
+
+	mux.Handle("/login", page.Login(*signInClient, appPublicURL, clientID, signInPublicURL))
+	mux.Handle("/home", page.Home(tmpls.Get("home.gohtml"), fmt.Sprintf("%s/login", appPublicURL), bundle.For("en"), page.En))
+	mux.Handle("/auth/callback", page.SetToken(*signInClient, appPublicURL, clientID, RandomString(12)))
```
hawx, line 85: we'll want that value to be different each time, so could pass the func in and let `SetToken` call it

app/internal/page/set_token.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"net/url"
+
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/signin"
+)
+
+func SetToken(c signin.Client, appPublicURL, clientID, JTI string) http.HandlerFunc {
```
hawx, line 12: `Callback` or `SigninCallback` or `AuthCallback` might be better names for this

app/internal/page/login.go
```diff
@@ -0,0 +1,18 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/signin"
+)
+
+func Login(c signin.Client, appPublicURL, clientID, signInBaseURL string) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		redirectURL := fmt.Sprintf("%s%s", appPublicURL, c.AuthCallbackPath)
```
hawx, line 12: might be nicer to construct the `redirectURL` in `main.go` then pass it down to the places needed, would also allow you to remove the `AuthCallbackPath` from `Client`

app/internal/signin/get_token.go
```diff
@@ -0,0 +1,119 @@
...
+		"jti": JTI,
+		"iat": time.Now().Unix(),
+	})
+
+	privateKey, err := c.secretsClient.PrivateKey()
+	if err != nil {
+		return "", err
+	}
+
+	signedAssertion := signJwt(string(data), privateKey)
```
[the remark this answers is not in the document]
hawx: also just noticed this isn't using the latest `v4` version of the package

app/internal/secrets/secrets.go
```diff
@@ -0,0 +1,93 @@
...
+	sess, err := session.NewSession(config)
+
+	if err != nil {
+		return nil, fmt.Errorf("error initialising new AWS session: %v", err)
+	}
+
+	svc := secretsmanager.New(
+		sess,
+		aws.NewConfig().WithRegion("eu-west-1"),
+	)
```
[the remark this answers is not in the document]
acsauk: Agree with the NewClient approach to drop baseURL from Client but on past project I *think* we relied on the credential provider chain in prod to get these from somewhere rather than having to manually set them. I've had to include the creds line as Localstack makes you set them to any value in order to work (or I guess we could also set as dummy values in env vars in our docker images too?)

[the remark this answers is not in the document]
hawx: yeah setting the env vars in docker-compose sounds smart, anything that makes it more like it will be in prod is good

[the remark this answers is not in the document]
acsauk: Yup, works as env vars so have dropped the creds line

app/internal/page/set_token.go
acsauk: `SigninCallback` ✅

app/main.go
acsauk: I've moved this to its own package as we'll probably get some re-use out of it (or at least have the place to add different types of random funcs in) with `state` and `nonce`:

Screenshot 2022-08-25 at 11 52 42

So the API is like `random.String(12)`. Any thoughts on this? I know utility-type classes can sometimes divide the room.

hawx: I guess that works, although you could also move it to `signin` since that is the only thing using it at the moment

hawx (approved):

andrewpearce-digital (approved):

PR #46, opened 2022-08-25
MLPAB-92 Add who is the lpa for page

hawx:


app/internal/page/paths.go
```diff
@@ -0,0 +1,8 @@
+package page
+
+const (
+	startPath          = "/"
+	donorDetailsPath   = "/donor-details"
+	donorAddressPath   = "/donor-address"
+	whoIsTheLpaForPath = "/who-is-the-lpa-for"
+)
```
hawx, line 8: I thought it would be nice to put these somewhere so they can be easily changed later, I've left them as strings for now but can easily be typed up if we have the need

johnnolan (approved):

PR #91, opened 2022-09-07
MLPAB-103 Add certificate provider details page

hawx:


app/internal/page/certificate_provider_details_test.go
```diff
@@ -0,0 +1,320 @@
...
+		"date-of-birth-year":  {"1990"},
+	}
+
+	r, _ := http.NewRequest(http.MethodPost, "/", strings.NewReader(form.Encode()))
+	r.Header.Add("Content-Type", formUrlEncoded)
+
+	err := CertificateProviderDetails(nil, dataStore)(appData, w, r)
+
+	assert.Equal(t, expectedError, err)
+	mock.AssertExpectationsForObjects(t, dataStore)
```
acsauk, line 190: Should we be checking for some kind of error status code here? Or are we holding this back for when we work on non-happy path?

acsauk (approved):
One Q on status codes for page errors but I'm assuming we're ignoring them for now.

:shipit:

app/internal/page/certificate_provider_details_test.go
hawx: Because it's returning an error the thing above will set it to 500, but in this test it would still be 200 which would be confusing to check (I think). But like you say, this will all change when we actually figure out what we're going to do for non-happy path stuff

PR #113, opened 2022-09-16
MLPAB-255 - Manage environment configuration for second region

andrewpearce-digital:
# Purpose

We want to be ready to deploy to a second region

Fixes MLPAB-255

## Approach

- Add an input variable to define regions we should deploy to
- Define eu-west-1 as a region we deploy to
- Define a second region instance of the region module (without requiring the deployment there)
- Add run-book explaining second region procedure

## Learning

- https://www.terraform.io/language/modules/develop/refactoring#moved-block-syntax

## Checklist

* ~I have added relevant logging with appropriate levels to my code~
* ~I have updated documentation (Confluence/GitHub wiki/tech debt doc) where relevant~
* ~I have added tests to prove my work~
* ~I have added welsh translation tags and updated translation files~
* ~I have run an accessibility tool on any pages I have made changes to and fixed any issues found~
* ~The product team have tested these changes~

### DIFF
--- a/docs/runbooks/deployment_into_additional_regions.md
+++ b/docs/runbooks/deployment_into_additional_regions.md
@@ -12,7 +12,7 @@ The terraform configuration includes a definition for a London region that can b
 
 Raise a pull request to set `region_replica_enabled` and `stream_enabled` to `true` if not already set to true.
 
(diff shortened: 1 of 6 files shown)


acsauk (approved):
Nice - looks like this will be super slick when/if we need it.

PR #119, opened 2022-09-26
MLPAB-137: Handle redirect from govuk pay

acsauk:
# Purpose

Adds the route to handle redirects from GOV.UK Pay in order to confirm the payment was successful and persist a payment reference against payment ID on the users LPA.

Fixes MLPAB-137

## Checklist


app/internal/page/payment_confirmation.go
```diff
@@ -0,0 +1,75 @@
...
+
+		paymentId := payCookie.Values[pay.CookiePaymentIdValueKey].(string)
+		getPaymentResponse, err := client.GetPayment(paymentId)
+
+		if err != nil {
+			logger.Print(fmt.Sprintf("unable to retrieve payment info: %s", err.Error()))
+			return err
+		}
+
+		paymentReference := random.String(12)
```
acsauk, line 44: This is a stop-gap until we decide on the format of our references - well aware this won't create truly unique strings.

app/internal/random/random.go
```diff
@@ -4,7 +4,13 @@ import (
 	"crypto/rand"
 )
 
-const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
+const charset = "abcdefghijklmnpqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ123456789"
+
+type RandomGenerator interface {
+	String(length int) string
+}
```
acsauk, line 11: So we can mock out randomness

app/internal/pay/client.go
```diff
@@ -21,6 +23,32 @@ type Client struct {
...
+	s := strings.Trim(string(b), "\"")
+	s = strings.Trim(s, "0Z")
+
+	t, err := time.Parse(time.RFC3339, s+"00Z")
+	if err != nil {
+		return err
+	}
+
+	*g = GovUKPayTime(t)
+	return nil
```
acsauk, line 38: This is really gross - I couldn't work out how to get Go to parse a date string in the format`2016-01-21T17:15:000Z`. `RFC3339` was closest but still slightly out.

app/internal/page/payment_confirmation_test.go
```diff
@@ -0,0 +1,280 @@
...
+		template := &mockTemplate{}
+		template.
+			On("Func", w, &paymentConfirmationData{App: appData, PaymentReference: "123456789012"}).
+			Return(nil)
+
+		r, _ := http.NewRequest(http.MethodGet, "/payment-confirmation", nil)
+
+		sessionsStore := &mockSessionsStore{}
+		ensureSessionHasPayCookie(sessionsStore, r)
+		ensureCookieIsExpiredInSession(sessionsStore, r, w)
```
acsauk, line 42: Not sure if this makes intent clearer in tests or muddys the water a bit - open to discussion on what others think.

terraform/environment/terraform.tfvars.json
```diff
@@ -9,7 +9,8 @@
       ],
       "app": {
         "env": {
-          "app_public_url": "https://opg-lpa-fd-prototype.apps.live.cloud-platform.service.justice.gov.uk"
+          "app_public_url": "",
```
acsauk, line 12: This needs a rethink but didn't want to bloat this PR any further

app/internal/pay/client.go
hawx: In the Pay docs they say it is like `2022-03-25T13:11:29.019Z`, which would be `time.RFC3339Nano`, so I think you can simplify this to

```go

func (g *GovUKPayTime) UnmarshalText(b []byte) error {
	t, err := time.Parse(time.RFC3339Nano, string(b))
	if err != nil {
		return err
	}

	*g = GovUKPayTime(t)
	return nil
}

func (g GovUKPayTime) MarshalText() ([]byte, error) {
	return []byte(g.Format(time.RFC3339Nano)), nil
}

func (g GovUKPayTime) Format(s string) string {
	return time.Time(g).Format(s)
}
```

(also using Un/MarshalText is easier to deal with since you are inside the string, I think the only time you'd not want to use it is when you have to deal with weird null behaviour, :eyes: Sirius...)

hawx: It would also be nice if we the structs exposed from this package used normal `time.Time` and we translated in the package, but happy for you to leave if it seems like too much hassle right now

app/internal/pay/client_test.go
```diff
@@ -295,6 +294,23 @@ func TestGetPayment(t *testing.T) {
...
+	t.Run("Returns an error if unable to make request", func(t *testing.T) {
+		server := httptest.NewServer(http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {}))
+
+		defer server.Close()
+
+		payClient := Client{BaseURL: "not an url", ApiKey: apiToken, HttpClient: server.Client()}
+
+		actualGPResponse, err := payClient.GetPayment(paymentId)
+
+		if err == nil {
```
hawx, line 306: you can use `if assert.NotNil(t, err) { ... }` when you want to guard other assertions behind another

hawx: or use `assert.ErrorContains(t, err, "the error message")`

app/internal/pay/client_test.go
```diff
@@ -295,6 +294,23 @@ func TestGetPayment(t *testing.T) {
...
+			t.Fatal("Expected an error but received nil")
+		}
+
+		assert.Contains(t, err.Error(), "unsupported protocol scheme")
+		assert.Equal(t, GetPaymentResponse{}, actualGPResponse, "Return value did not match")
```
hawx, line 311: I don't really think this is necessary because the error would be handled so you'd never look at the response

app/internal/pay/data.go
```diff
@@ -1,6 +1,11 @@
 package pay
 
-import "time"
+const (
+	PaymentStatusSuccess    = "success"
```
hawx, line 4: unused?

app/internal/random/random.go
hawx: Unnecessary since you can have functions take a `func(int) string` argument and pass in `random.String`. Take a look at `page/login.go` and how it is injected in `main.go`

app/web/template/payment_confirmation.gohtml
```diff
@@ -0,0 +1,22 @@
...
+                <div class="govuk-panel__body">
+                    {{ tr .App "yourRefNumber" }}<br><strong>{{ .PaymentReference }}</strong>
+                </div>
+            </div>
+
+            <h2 class="govuk-heading-m">{{ tr .App "thankYouHeader" }}</h2>
+
+            {{ trHtml .App "paymentConfirmationContent" }}
+
+            <a class="govuk-button" href="/task-list">{{ tr .App "continueLink" }}</a>
```
hawx, line 19: change the href to `{{ link .App "/task-list" }}` so it handles welsh

terraform/environment/region/modules/app/ecs.tf
```diff
@@ -236,17 +236,21 @@ locals {
...
         },
         {
-          # this is not the final value, but will allow signin to be tested while the real redirectURL is changed
           name  = "APP_PUBLIC_URL",
-          value = var.app_env_vars.app_public_url
+          value = var.app_env_vars.app_public_url == "" ? local.dev_app_fqdn : var.app_env_vars.app_public_url
+        },
+        {
+          # this is not the final value, but will allow signin to be tested while the real redirectURL is changed
+          name  = "AUTH_REDIRECT_BASE_URL",
```
hawx, line 244: what is this doing?

terraform/environment/terraform.tfvars.json
hawx: could you revert all of the changes you've done to the terraform because auth no longer works

terraform/environment/region/modules/app/ecs.tf
acsauk: This was added as I need a reference to the application base URL to build a return URL for pay to redirect the user back to. As this was set to the prototype for dev envs it means the pay flow breaks after a user completes their payment info. 

Aaaand I just noticed I didn't push the changes to main that actually use this value 🤦

app/internal/random/random.go
acsauk: Nice, didn't realise this was a thing

app/internal/page/payment_confirmation_test.go
acsauk: That definitely reads better and feels nice to group on the mock object to keep context clear and reduce arguments - I'll update.

terraform/environment/terraform.tfvars.json
```diff
@@ -69,7 +71,8 @@
       ],
       "app": {
         "env": {
-          "app_public_url": "https://app.modernising.opg.service.justice.gov.uk"
+          "app_public_url": "https://app.modernising.opg.service.justice.gov.uk",
+          "auth_redirect_base_url": "https://opg-lpa-fd-prototype.apps.live.cloud-platform.service.justice.gov.uk"
```
hawx, line 75: I see, so this and the one for preprod should be the same as `app_public_url`

app/internal/page/about_payment.go
```diff
@@ -55,19 +54,18 @@ func AboutPayment(logger Logger, tmpl template.Template, sessionStore sessions.S
 				Secure:   secureCookies,
 			}
 
-			session := sessions.NewSession(sessionStore, "pay")
+			session := sessions.NewSession(sessionStore, pay.CookieName)
```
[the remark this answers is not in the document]
acsauk: My thinking was it was tied in with pay-specific cookies but now I see that written down it does seem like a blur of the boundaries of responsibilities. We have the concept of data in a few of our internal packages - how do you feel about having this in our main package to start to extract some of our magic strings and values out to?

app/internal/page/about_payment.go
```diff
@@ -26,7 +26,7 @@ func AboutPayment(logger Logger, tmpl template.Template, sessionStore sessions.S
 
 		if r.Method == http.MethodPost {
 			createPaymentBody := pay.CreatePaymentBody{
```
hawx, line 28: Have you seen https://docs.payments.service.gov.uk/optional_features/welsh_language/?

app/internal/page/about_payment.go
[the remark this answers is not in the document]
hawx: I'm fine with this being a constant, I just think it should stay where it is needed in the `page` module. The `pay` module (in my mind) should be an API client, it shouldn't have anything to do with cookies or sessions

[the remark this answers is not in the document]
acsauk: Fair point - `page` it is!

app/internal/page/payment_confirmation.go
```diff
@@ -0,0 +1,75 @@
...
+		lpa.PaymentDetails = PaymentDetails{
+			PaymentReference: paymentReference,
```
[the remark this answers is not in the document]
acsauk: Yeah you're right - must have missed that in the docs.

app/web/template/payment_confirmation.gohtml
acsauk: This is handy 👍

app/internal/page/about_payment.go
acsauk: Perfect - I've updated this now to define welsh for the language setting and return URL.

acsauk:
I've opened a PR in the GOV UK Pay repo https://github.com/alphagov/pay-publicapi/pull/1575 to see if we can get the date format updated to match the actual format returned by their API. Until that's updated our cypress tests are going to fail due to the discrepancy in date format 😞

app/internal/page/payment_confirmation_test.go
```diff
@@ -0,0 +1,232 @@
...
+	})
+}
+
+func (m *mockDataStore) withLpaDataInStore() *mockDataStore {
+	m.On("Get", mock.Anything, "session-id").Return(nil)
+
+	return m
+}
+
+func (m *mockDataStore) withUpdatedLpaData(paymentId, paymentReference string) *mockDataStore {
```
hawx, line 169: given this method is available in all the tests it might be better to give it a more specific name

app/internal/page/payment_confirmation_test.go
```diff
@@ -0,0 +1,232 @@
...
+		err := PaymentConfirmation(logger, template.Func, payClient, dataStore, sessionsStore)(appData, w, r)
+		resp := w.Result()
+
+		assert.Nil(t, err)
+		assert.Equal(t, http.StatusOK, resp.StatusCode)
+		mock.AssertExpectationsForObjects(t, dataStore, sessionsStore, logger, payClient)
+	})
+}
+
+func (m *mockDataStore) withLpaDataInStore() *mockDataStore {
```
hawx, line 163: because of how the mockDataStore works this method name is weird, it is only saying that a `Get` call will not error not that there is any data in the store, as you'd set that like `&mockDataStore{data: myLpaData}`

app/internal/pay/client.go
```diff
@@ -55,11 +77,17 @@ func (c *Client) GetPayment(paymentId string) (GetPaymentResponse, error) {
 		return GetPaymentResponse{}, err
 	}
 
+	fmt.Println(c.BaseURL + "/v1/payments/" + paymentId)
```
hawx, line 80: did you want to keep this in?

app/internal/page/app.go
```diff
@@ -18,16 +18,34 @@ import (
...
 	if l == En {
-		http.Redirect(w, r, url, code)
+		return EnglishAbbreviation, nil
+	}
+
+	if l == Cy {
+		return WelshAbbreviation, nil
+	}
+
+	return "", fmt.Errorf("unsupported language '%v'", l)
```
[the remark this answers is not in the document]
acsauk: Good call - it was bugging me having to handle an error here but a default seems like a good compromise.

app/internal/page/app.go
```diff
@@ -18,16 +18,34 @@ import (
...
-		http.Redirect(w, r, "/cy"+url, code)
+		return url
 	}
 }
 
 const (
 	En Lang = iota
 	Cy
+	EnglishAbbreviation = "en"
+	WelshAbbreviation   = "cy"
```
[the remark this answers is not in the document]
acsauk: Yup, will demote to private.

app/internal/page/payment_confirmation_test.go
acsauk: Was struggling with this name - I'll have a play around with some options.

app/internal/pay/client.go
acsauk: Nope, thanks!

hawx (approved):

PR #121, opened 2022-09-28
MLPAB-240 Add guidance pages

hawx:


cypress/e2e/what-happens-next.cy.js
```diff
@@ -0,0 +1,14 @@
...
+        cy.visit('/testing-start?redirect=/what-happens-next');
+
+        cy.injectAxe();
+        cy.checkA11y(null, {
+            rules: { region: { enabled: false } },
+        });
+
+        cy.contains('a', 'Continue').click();
+
+        cy.url().should('contain', '/what-happens-when-signing');
```
hawx, line 12: these tests are weak, once the whole flow is in we should probably replace with one that just clicks through all the signing content

johnnolan (approved):

PR #122, opened 2022-09-29
MLPAB-184 Add 'read and sign your lpa' page

hawx:


cypress/e2e/read-your-lpa.cy.js
```diff
@@ -0,0 +1,25 @@
+describe('Check the LPA', () => {
```
acsauk, line 1: Should this be:

```suggestion
describe('Read your LPA', () => {
```
Or is this section a continuation of the check?

acsauk (approved):
:shipit:

PR #128, opened 2022-10-03
MLPAB-166 Add Yoti integration

hawx:
Locally the client will return fake data, on PR envs it will hit the sandbox, and in preprod/prod it will use the real thing.


.codecov.yml
```diff
@@ -5,6 +5,7 @@ coverage:
   ignore:
     - "*_test.go"
     - "./mocks/*"
+    - "./app/internal/easyid"
```
hawx, line 8: because the SDK makes things private the only way to test it would be to mock the server side (which would be very annoying) or write our own implementation

terraform/environment/terraform.tfvars.json
```diff
@@ -10,7 +10,10 @@
       "app": {
         "env": {
           "app_public_url": "",
-          "auth_redirect_base_url": "https://opg-lpa-fd-prototype.apps.live.cloud-platform.service.justice.gov.uk"
+            "auth_redirect_base_url": "https://opg-lpa-fd-prototype.apps.live.cloud-platform.service.justice.gov.uk",
```
andrewpearce-digital, line 13: There's an extra indentation on these that shouldn't be here

andrewpearce-digital (approved):

PR #135, opened 2022-10-05
MLPAB-243 Yoti and general ID flow

hawx:


app/internal/page/data.go
```diff
@@ -78,7 +81,14 @@ type Lpa struct {
...
 	SignatureCode            string
-	IdentityOptions          []IdentityOption
+	IdentityOptions          IdentityOptions
+}
+
+type IdentityOptions struct {
+	Selected []IdentityOption
+	First    IdentityOption
+	Second   IdentityOption
+	Current  int
```
hawx, line 91: probably would be better to store the results then confirm that the data is present to move on, also need to figure out what happens if you leave half-way through then start again from the task list

hawx: yeah, going to change this so there are `FirstDone`/`SecondDone` flags that get set by the callback route (not the redirect), then can skip to the callback if the flag is set and the continue click won't double set

hawx: and then can probably move the logic into handlers instead of redirecting multiple times

andrewpearce-digital (approved):

PR #142, opened 2022-10-07
MLPAB-276 - Mange regional default VPC default security group EC2.2

andrewpearce-digital:
# Purpose

We do not recommend using the default security group. Because the default security group cannot be deleted, you should change the default security group rules setting to restrict inbound and outbound traffic. This prevents unintended traffic if the default security group is accidentally configured for resources such as EC2 instances.

Fixes MLPAB-276

## Approach

- Identify the default VPC with a data source
- Remove ingress and egress rules from the default VPC default security group

Prior to merging we will import the default security group for management in each account terraform workspace

```shell
tf workspace select <workspace_name>
tf import 'module.eu_west_1[0].aws_default_security_group.default' <security_group_id>
```

## Learning

- https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#ec2-2-remediation
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_security_group
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/vpc#default

## Checklist


acsauk (approved):
:shipit:

PR #151, opened 2022-10-11
MLAPB-308: Add gitleaks baseline generator command and ignore baseline

acsauk:
# Purpose
We're no longer scanning the baseline file with this change which should lead to a slimmer baseline file. I've also added a make command to generate a fresh baseline file for when we accept dummy/false positives in code going forward.

Fixes MLPAB-308

### DIFF
--- a/.gitleaks.toml
+++ b/.gitleaks.toml
@@ -8,7 +8,7 @@ useDefault = true
 [allowlist]
 description = "global allow lists"
 paths = [
-    '''gitleaks.toml''',
+    '''.gitleaks.toml''',
     '''(.*?)(jpg|gif|doc|docx|zip|xls|pdf|bin|svg|socket)$''',
     '''(go.mod|go.sum)$''',
     '''node_modules''',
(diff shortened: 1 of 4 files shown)


.gitleaks.toml
```diff
@@ -0,0 +1,16 @@
...
+
+[extend]
+# useDefault will extend the base configuration with the default gitleaks config:
+# https://github.com/zricethezav/gitleaks/blob/master/config/gitleaks.toml
+useDefault = true
+
+[allowlist]
+description = "global allow lists"
+paths = [
+    '''gitleaks.toml''',
```
andrewpearce-digital, line 11: should this be `.gitleaks.toml`? 
```suggestion
    '''.gitleaks.toml''',
```

acsauk: Technically yes but I think its a fuzzy match. Will update to stop any eye twitching in the future 😁

andrewpearce-digital (dismissed):

andrewpearce-digital (approved):

PR #164, opened 2022-10-18
MLPAB-326 Add validation for donor details and address

hawx:


app/internal/page/your_address_test.go
```diff
@@ -672,9 +671,38 @@ func TestYourAddressFormValidate(t *testing.T) {
...
-				"address-postcode": "enterPostcode",
+				"address-line-1": "enterAddress",
+				"address-town":   "enterTownOrCity",
+			},
+		},
+		"manual-max-length": {
+			form: &yourAddressForm{
+				Action: "manual",
+				Address: &place.Address{
+					Line1:      strings.Repeat("x", 50),
```
acsauk, line 682: This is handy!

app/internal/page/your_details.go
```diff
@@ -86,21 +96,51 @@ func (d *yourDetailsForm) Validate() map[string]string {
 	if d.FirstNames == "" {
 		errors["first-names"] = "enterFirstNames"
 	}
+	if len(d.FirstNames) > 53 {
+		errors["first-names"] = "firstNamesTooLong"
+	}
```
acsauk, line 101: Probably not something we can resolve on this PR but do we have the research/reasoning on these specific limits on name lengths? Feels like a hangover from needing to print to a paper form.

acsauk (approved):
One Q on where the name length limits have come from but we can clear that up as a team if you don't have the info.

app/internal/page/your_details.go
hawx: I'm just taking them from what was in the spreadsheet shared from Make

PR #177, opened 2022-10-26
MLPAB-83 Try adding aws xray stuff

hawx:


terraform/environment/region/modules/app/ecs.tf
```diff
@@ -290,8 +290,35 @@ locals {
...
+        },
+        {
+          name  = "XRAY_ENABLED",
+          value = "1"
         }
       ]
     }
   )
+
+  aws_otel_collector = jsonencode(
```
andrewpearce-digital, line 302: when you've got this working, could you add `readonlyRootFilesystem = true` to this container definition please. If it fails to work with that configured, let me know and I'll suppress the finding in AWS SecurityHub

.codecov.yml
```diff
@@ -6,6 +6,7 @@ coverage:
     - "*_test.go"
     - "./mocks/*"
     - "./app/internal/identity/yoti*"
+    - "./app/internal/telemetry"
```
hawx, line 9: maybe in the future it will contain actual code, but for now tests aren't useful (in my opinion)

acsauk (approved):
:shipit: will be interested to try this out - I've heard lots about x-ray but never used it first hand.

PR #181, opened 2022-10-27
MLPAB-373 - Amazon EC2 should be configured to use VPC endpoints that are created for the Amazon EC2 service

andrewpearce-digital:
# Purpose

To improve the security posture of your VPC, you can configure Amazon EC2 to use an interface VPC endpoint. Interface endpoints are powered by AWS PrivateLink, a technology that enables you to access Amazon EC2 API operations privately. It restricts all network traffic between your VPC and Amazon EC2 to the Amazon network. Because endpoints are supported within the same Region only, you cannot create an endpoint between a VPC and a service in a different Region. This prevents unintended Amazon EC2 API calls to other Regions. 

Fixes MLPAB-373

## Approach

- Configure an EC2 VPC endpoint for the default VPC

## Learning

- https://docs.aws.amazon.com/securityhub/latest/userguide/securityhub-standards-fsbp-controls.html#ec2-10-remediation

## Checklist

### DIFF
--- a/terraform/account/region/default_vpc.tf
+++ b/terraform/account/region/default_vpc.tf
@@ -209,7 +209,7 @@ resource "aws_vpc_endpoint" "default_private" {
   vpc_endpoint_type   = "Interface"
   private_dns_enabled = true
   security_group_ids  = aws_security_group.default_vpc_endpoints[*].id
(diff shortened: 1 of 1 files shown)


acsauk (approved):
:shipit:

PR #195, opened 2022-11-08
MLPAB-354: Do you want replacement attorneys?

acsauk:
# Purpose

Updates the first page in the replacement attorney flow with warnings (content will be updated once Uchenna is back in)

Fixes MLPAB-354.

## Approach

Just getting some conditional warnings and initial plumbing on the first page to keep the next PR small.

## Checklist

* [ ] I have performed a self-review of my own code


app/internal/page/app.go
```diff
@@ -243,6 +244,23 @@ func testingStart(store sessions.Store, lpaStore LpaStore) http.HandlerFunc {
...
+			case "jointly-and-severally":
+				lpa.DecisionsType = "jointly-and-severally"
+			default:
+				lpa.DecisionsType = "mixed"
+				lpa.DecisionsDetails = "some details"
+			}
+
+			_ = lpaStore.Put(r.Context(), sessionID, lpa)
+		}
+
```
acsauk, line 263: After the replacement attorney flow is completed we'll work out how we're handling fixtures for cypress - definitely time with this addition.

app/internal/page/app_test.go
```diff
@@ -333,6 +333,55 @@ func TestTestingStart(t *testing.T) {
 		assert.Equal(t, "/somewhere", resp.Header.Get("Location"))
 		mock.AssertExpectationsForObjects(t, sessionsStore, lpaStore)
 	})
+
+	t.Run("how attorneys act", func(t *testing.T) {
```
acsauk, line 337: @hawx - I'm pretty sure this test passes no matter what the props on the LPA are set to. I'd expect `AssertExpectationsForObjects()` to trigger a failure if `lpa.DecisionType` doesn't match but it passes anyway. Any idea what I'm missing? This applies to the other tests in this series too after I tried to break some of them.

hawx: I think the problem is that the `Get` call will return `lpa` (which is a pointer, so the real code will mutate the value), then the `Put` mock will obviously have the right values because the code has just changed them (so your setting of `DecisionsType` will be overwritten).

If you inline `lpa` you'll see the test can be broken:

```go

lpaStore := &mockLpaStore{}
lpaStore.
	On("Get", ctx, mock.Anything).
	Return(&Lpa{}, nil)

lpaStore.
	On("Put", ctx, mock.Anything, &Lpa{DecisionsType: tc.DecisionsType, DecisionsDetails: "what"}). // e.g. for the failure
	Return(nil)
```

acsauk: Pointers again 😂 cheers for that, will update!

app/web/template/want_replacement_attorneys.gohtml
```diff
@@ -7,19 +7,32 @@
...
-        {{ trHtml .App "wantReplacementAttorneysContent" }}
+        <div>
+          {{ if eq .Lpa.DecisionsType "jointly" }}
+            {{ template "warning" (warning .App "jointlyReplacementWarning") }}
+            {{ trHtml .App "jointlyReplacementWarningContent" }}
+          {{ end }}
 
-        <div class="govuk-form-group {{ if index .Errors "want" }}govuk-form-group--error{{ end }}">
-          {{ template "error-message" (errorMessage . "want") }}
+          {{ if eq .Lpa.DecisionsType "mixed" }}
```
hawx, line 18: you can use `{{ else if ... }}` in templates too

app/internal/page/app_test.go
hawx: Yep, I'd just recommend writing things out in full in the tests because mutating can introduce subtle issues like this

hawx (approved):

PR #197, opened 2022-11-10
MLPAB-421 - Create a KMS key for S3 buckets

andrewpearce-digital:
# Purpose

Use customer managed encryption key for S3 buckets.

Fixes MLPAB-421

## Approach

- Create an encryption key for S3 and pass it into the region module

## Learning

- https://docs.aws.amazon.com/AmazonS3/latest/userguide/specifying-kms-encryption.html

## Checklist

### DIFF
--- a/terraform/account/region/data_sources.tf
+++ b/terraform/account/region/data_sources.tf
@@ -2,3 +2,8 @@ data "aws_kms_alias" "cloudwatch_application_logs_encryption" {
   name     = var.cloudwatch_log_group_kms_key_alias
   provider = aws.region
 }
+
+# data "aws_kms_alias" "s3_encryption" {
+#   name     = var.s3_kms_key_alias
+#   provider = aws.region
+# }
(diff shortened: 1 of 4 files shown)


terraform/account/region/data_sources.tf
```diff
@@ -2,3 +2,8 @@ data "aws_kms_alias" "cloudwatch_application_logs_encryption" {
   name     = var.cloudwatch_log_group_kms_key_alias
   provider = aws.region
 }
+
+# data "aws_kms_alias" "s3_encryption" {
+#   name     = var.s3_kms_key_alias
+#   provider = aws.region
+# }
```
acsauk, line 9: Are we saving this for a future ticket?

acsauk (approved):
:shipit: with one Q

terraform/account/region/data_sources.tf
andrewpearce-digital: Yeah, the next ticket will use this to encrypt the access logs bucket

PR #199, opened 2022-11-11
MLPAB-428 - Enable bucket keys encryption with CMK

andrewpearce-digital:
# Purpose

Use an AWS KMS customer managed key (CMK) when encrypting S3 buckets, and use bucket keys to reduce AWS KMS usage.

Fixes MLPAB-428

## Approach

- Retrieve key ARN using data source
- Use CMK to encrypt bucket
- Enable Bucket Keys
- Require server side encryption when putting objects

## Learning

- https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html#require-sse-kms
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-key.html

## Checklist

### DIFF
--- a/terraform/account/region/data_sources.tf
+++ b/terraform/account/region/data_sources.tf
@@ -3,7 +3,7 @@ data "aws_kms_alias" "cloudwatch_application_logs_encryption" {
   provider = aws.region
 }
 
-# data "aws_kms_alias" "s3_encryption" {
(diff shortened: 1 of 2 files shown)


acsauk (approved):
:shipit:

PR #200, opened 2022-11-11
MLPAB-429 - Disable bucket key and CMK Encryption

andrewpearce-digital:
# Purpose

I believe I'm unable at this time to configure the encryption context required to use bucket keys for ALB access logs

Fixes MLPAB-428

## Approach

- Set bucket key to false

## Learning

- https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingKMSEncryption.html#encryption-context

## Checklist


acsauk (approved):
:shipit:

PR #205, opened 2022-11-14
MLPAB-406: How should replacement attorneys step in

acsauk:
# Purpose

Adds the final page to the replacement attorney flow.

Fixes MLPAB-406

## Approach

I've purposefully left out updating task list status here ready to tackle in a separate ticket.

## Learning

JSON keys are case insensitive 🙃 spent about an hour trying to work out why I had valid JSON but the translations wouldn't compile. As far as JSON/translation package is concerned, `Other: ""` === `other: ""`.

## Checklist


app/internal/page/how_should_replacement_attorneys_step_in.go
```diff
@@ -9,16 +9,84 @@ import (
...
+			if len(data.Errors) == 0 {
+				lpa.HowShouldReplacementAttorneysStepIn = data.Form.WhenToStepIn
+
+				if data.Form.WhenToStepIn != "other" {
+					lpa.HowShouldReplacementAttorneysStepInDetails = ""
+				} else {
+					lpa.HowShouldReplacementAttorneysStepInDetails = data.Form.OtherDetails
+				}
+
+				lpa.HowShouldReplacementAttorneysStepInDetails = data.Form.OtherDetails
```
hawx, line 48: I guess you meant to delete the `if-else` above?

app/web/template/how_should_replacement_attorneys_step_in.gohtml
```diff
@@ -3,7 +3,62 @@
 {{ define "main" }}
```
hawx, line 3: image

some of these need to be `trHtml`

cypress/e2e/how-should-replacement-attorneys-step-in.cy.js
```diff
@@ -0,0 +1,39 @@
+describe('How should replacement attorneys step in', () => {
```
hawx, line 1: I think it would be cool after this to rework all of the attorney tests to do the journey each time for each scenario, just because the pages are very interlinked in specific ways.

(And we have weird edge cases that haven't really been thought of; like going through with one attorney and then going back to add another could invalidate what was selected later, but you'd be able to skip back to the task list and leave it in a weird state)

acsauk: Totally agree - once we've got a really clear idea of the different states of the section and the business logic behind it we can merge a bunch of these to give us some confidence.

app/web/template/how_should_replacement_attorneys_step_in.gohtml
acsauk: Doh, thanks for catching.

app/internal/page/how_should_replacement_attorneys_step_in.go
acsauk: The intent here was for people coming back and changing their response to include something other than `other`. I think what needs to happen is dropping line 48 so we only set the details when the response is `other`. This will avoid persisting `OtherDetails` that's prepopulated in the form when they first hit the page.

hawx: I thought because of the way the validation works you'd only need line 48 since selecting other validates you must have details, selecting non-other validates details is empty.

hawx (approved):

PR #212, opened 2022-11-28
MLPAB-468: Content and flow updates

acsauk:
# Purpose

This is a bunch of ad hoc content and redirect updates following a team run-through of the app in preparation for UR. This also updates the back button to return to the previous page rather than the task list.

Fixes MLPAB-429, MLPAB-432, MLPAB-468, MLPAB-469, MLPAB-470, MLPAB-471, MLPAB-472

## Approach

Due to some of the content changes that rely on pluralism and some templated text I've added `trFormatCount()`. Not sure how we all feel about this but I couldn't see a way to utilise the count and format funcs separately to achieve what `trFormatCount()` is doing.

## Checklist


app/internal/page/app.go
```diff
@@ -346,6 +347,7 @@ func makeHandle(mux *http.ServeMux, logger Logger, store sessions.Store, localiz
 				CookieConsentSet: cookieErr != http.ErrNoCookie,
 				CanGoBack:        opt&CanGoBack != 0,
 				RumConfig:        rumConfig,
+				PreviousPath:     r.Header.Get("Referer"),
```
hawx, line 350: this just creates a loop between 2 pages, maybe it should be done in JS so it uses the browser history (because I would hate us to have to store a list server side to play back) _or_ I now think we should remove the back button completely because it will be very annoying/difficult to get it to work in a logical way, and we can come back to it later if/when we have more research supporting us adding one

app/internal/templatefn/fn.go
```diff
@@ -37,6 +39,15 @@ var All = map[string]interface{}{
...
 }
 
+var LpaPropertyToTransKey = map[string]string{
+	"pfa":                   "lpaTypePfa",
+	"jointly":               "jointly",
+	"jointly-and-severally": "jointlyAndSeverally",
+	"mixed":                 "jointlyAndSeverallyMixed",
+	"when-registered":       "asSoonAsItsRegistered",
+	"when-capacity-lost":    "onlyWhenIDoNotHaveMentalCapacity",
+}
```
hawx, line 49: why not just make these normal translations?

app/internal/page/app.go
acsauk: 😂 I knew there was going to be some flaw in this. I'd only tested going back a single page, making a change, and continuing.

I'll see what the appetite is for dropping this for UR in standup tomorrow - I agree other options are going to be complex for what seems like replicating the browser back button.

app/internal/templatefn/fn.go
acsauk: Mainly because my eye was twitching at the mix of styles in the JSON keys! Having to write that I now realise that's a pretty lame reason to add more complexity so I'll suck up my OCD and add some kebab case keys in the translation files.

hawx (approved):

PR #227, opened 2022-12-06
MLPAB-466: Group paths under a struct and use in app

acsauk:
# Purpose

Moves our app paths from consts to a struct and passes this around the app where required.

Fixes MLPAB-466

## Approach

The only real change to how anything works is injecting the Paths into the `identityOptions` struct and paths into `App()`. I've left a comment about my feelings on the approach for `identityOptions`.

## Checklist

### DIFF
--- a/app/internal/page/identity_options.go
+++ b/app/internal/page/identity_options.go
@@ -97,22 +97,23 @@ type IdentityOptions struct {
 	Selected []IdentityOption
 	First    IdentityOption
 	Second   IdentityOption
+	Paths    AppPaths
 }
 
 func (o IdentityOptions) NextPath(current IdentityOption) string {
 	identityOptionPaths := map[IdentityOption]string{
-		Yoti:                     identityWithYotiPath,
-		Passport:                 identityWithPassportPath,
-		DrivingLicence:           identityWithDrivingLicencePath,
-		GovernmentGatewayAccount: identityWithGovernmentGatewayAccountPath,
-		DwpAccount:               identityWithDwpAccountPath,
-		OnlineBankAccount:        identityWithOnlineBankAccountPath,
-		UtilityBill:              identityWithUtilityBillPath,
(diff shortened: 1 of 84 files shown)


app/internal/page/identity_options.go
```diff
@@ -97,22 +97,23 @@ type IdentityOptions struct {
 	Selected []IdentityOption
 	First    IdentityOption
 	Second   IdentityOption
+	Paths    AppPaths
```
acsauk, line 100: I'm a little bit iffy about this approach. Although it's going to be rare for our URLs to change there may be a need to do this in the future and then as we persist identity options on the Lpa struct we could have out-of-date paths. We could potentially rework the flow to query whatever the path should be as its needed but wanted to get some feedback first (and would also be a future ticket)

app/internal/page/task_list.go
```diff
@@ -6,6 +6,23 @@ import (
...
+	ChooseCertificateProviderTask         = "chooseYourCertificateProvider"
+	CheckAndSendToCertificateProviderTask = "checkAndSendToYourCertificateProvider"
+	PayForLpaSection                      = "payForTheLpa"
+	PayForTheLpaTask                      = "payForTheLpa"
+	ConfirmYourIdentityAndSignSection     = "confirmYourIdentityAndSign"
+	ConfirmYourIdentityAndSignTask        = "confirmYourIdentityAndSign"
+	RegisterTheLpaSection                 = "registerTheLpa"
+	RegisterTheLpaTask                    = "registerTheLpa"
+)
+
```
acsauk, line 25: Started this when I was considering grouping paths by section. Having worked through the ticket I don't think we need to (at the moment) but if we need to support a defined flow to be able to go back we could lean on the consts here in other places.

app/internal/page/identity_options.go
acsauk: Much better - actioned ✅

hawx (approved):

PR #230, opened 2022-12-12
MLPAB-78 - Only build and deploy new image if /app has changed

andrewpearce-digital:
# Purpose

In order to speed up PR workflow, only build and deploy a new app image when there are changes to /app.

Fixes MLPAB-78

## Approach

- check for changes to /app (and Dockerfile, package.json and yarn.lock) compared to main branch
  - if changes, output changes = true, create a new tag
  - if no changes, output changes = false, pull image tag on production
  - if no changes, Skip unit tests, docker build and ui tests on docker image
  - if no changes, Notify user via Step Summary that jobs are being skipped

image

## Learning

- https://carlosbecker.com/posts/git-changed/
- https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary

## Checklist


.github/workflows/docker_job.yml
```diff
@@ -36,12 +42,15 @@ jobs:
       - name: Upload Trivy scan results to GitHub Security tab
         id: trivy_upload_sarif
         uses: github/codeql-action/upload-sarif@v2
-        if: always()
+        if: inputs.changes_detected == true
```
jamesrwarren, line 45: you might need to do always() && inputs.changes_detected == true here. If you do this then it won't try and run if changes not detected but will still always run even on failure of previous job. If I remember rightly why it was set to always() in the first place.

andrewpearce-digital: that's gone back to always()

.github/workflows/tags_job.yml
```diff
@@ -55,3 +61,42 @@ jobs:
...
+        uses: aws-actions/configure-aws-credentials@v1.7.0
+        with:
+          aws-region: us-east-1
+          role-to-assume: arn:aws:iam::311462405659:role/modernising-lpa-github-actions-ssm-get-parameter
+          role-duration-seconds: 900
+          role-session-name: GithubActionsSSMGetParameter
+      - name: Pull production environment tag
+        id: pull_production_tag
+        run: |
+          value=$( aws ssm get-parameter --name '/modernising-lpa/container-version/preproduction' | jq -r .Parameter.Value )
```
hawx, line 79: pull production tag but gets preproduction?

andrewpearce-digital: we're not writing a production tag at the moment because it's not on the path to live.

hawx (approved):

hawx (approved):

PR #232, opened 2022-12-13
MLPAB-522 - Use docker compose pull instead of docker compose build for Cypress tests

andrewpearce-digital:
# Purpose

Optimise Cypress test workflow by using the previously built image instead of rebuilding it

Fixes MLPAB-522

## Approach

- Use docker compose pull
- Remove docker compose build

## Checklist


acsauk (approved):
:shipit:

PR #234, opened 2022-12-13
MLPAB-412: Allow donors to add, amend and remove people to notify

acsauk:
# Purpose

Given we have an existing pattern for CRUD-ing multiple 'things' I did the whole process in a single ticket here. 

Fixes MLPAB-412.

## Approach

Mirroring adding attorneys and replacement attorneys this change allows a donor to add multiple people to notify about the LPA. I've not put limits on the amount yet as it felt better to tackle that once we have design and content in place for it and you'll see there are some areas here where I've gone with boilerplate.

There are also a lot of very similar patterns in route handlers for adding multiple 'things' flow which may be ripe for a refactor to consolidate in one place at some point.

## Learning

I've tried a different approach with the cypress tests here relying on some fixture data that we can reference in tests and bringing some repetitive steps under a Cypress command. This also means we're testing a whole user flow for each test rather than pages in isolation which should hopefully give us some more confidence.

If we agree on fixtures being a thing for our project I'd like to enable seeding from them in some future tickets so we can replace our growing list of query params we append to `/test-start`. We'd need to be careful not to bloat them to an unusable level but it feels more appropriate to have the data defined in our testing domain rather than between the app and our tests.

## Checklist


app/web/template/layout/people-to-notify-summary.gohtml
```diff
@@ -0,0 +1,62 @@
+{{ define "people-to-notify-summary" }}
+    {{range $i, $p := .PeopleToNotify}}
+        {{ $personNumber := inc $i }}
+
+        <h2 class="govuk-heading-m">{{ trFormat $.App "personCount" "PersonNumber" $personNumber }}</h2>
```
hawx, line 5: this header looks weird, if we are keeping it then it should probably be smaller and lower priority than the "People to notify" section header

cypress/support/commands.js
```diff
@@ -48,3 +48,47 @@ function terminalLog(violations) {
 Cypress.Commands.add('checkA11yVvv', () => {
     cy.checkA11y(null, { rules: { region: { enabled: false } } }, terminalLog);
 })
+
+Cypress.Commands.add('addPersonToNotify', (p) => {
```
hawx, line 52: I think it would be better to have this as a function in the test file rather than a command, unless you plan on using it elsewhere?

acsauk: Potentially has further uses in the future but not right now so I'll move to where its used 👍

app/web/template/layout/people-to-notify-summary.gohtml
acsauk: @hawx are you referring to the physical font size with this or the hierarchical `h` elements? The markup looks to be in order:

Screenshot 2022-12-15 at 10 43 45

And the physical font size looks like its comparatively less prominent than the page header:

Screenshot 2022-12-15 at 10 43 52

hawx (approved):

PR #242, opened 2022-12-19
Return nil from Redirect

hawx:
seeing

```
appData.Lang.Redirect(...)
return nil
```

has been annoying me for a while


acsauk (approved):
Much better 👏

PR #247, opened 2022-12-20
MLPAB-508: Sign your LPA

acsauk:
# Purpose

Adds sign your LPA page as per prototype design.

Fixes MLPAB-508.

## Approach

The DonorSignature feels a little redundant right now but I needed the format to work with the checkbox template. 

## Checklist


app/internal/page/sign_your_lpa.go
```diff
@@ -0,0 +1,96 @@
...
+type signYourLpaData struct {
+	App    AppData
+	Errors map[string]string
+	Lpa    *Lpa
+	Form   *signYourLpaForm
+}
+
+const (
+	CertificateProviderHasWitnessed = "cp-witnessed"
+	WantToApplyForLpa               = "want-to-apply"
```
hawx, line 20: if you're going to pull these out then I think you should also inject them into the template to make the circle complete

app/internal/page/sign_your_lpa.go
```diff
@@ -0,0 +1,96 @@
...
+
+		return tmpl(w, data)
+	}
+}
+
+func readSignYourLpaForm(r *http.Request) *signYourLpaForm {
+	r.ParseForm()
+
+	form := &signYourLpaForm{
+		DonorSignatures: []string{},
```
hawx, line 68: why make this a list?

app/internal/page/sign_your_lpa.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+func (f *signYourLpaForm) Validate() map[string]string {
+	errors := map[string]string{}
+
+	if !slices.Contains(f.DonorSignatures, CertificateProviderHasWitnessed) {
+		errors["sign-lpa"] = "selectBothBoxes"
+	}
+
+	if !slices.Contains(f.DonorSignatures, WantToApplyForLpa) {
```
hawx, line 91: I would have understood making this a list if the validation was `len(f.DonorSignatures) == 2`

app/internal/page/sign_your_lpa_test.go
```diff
@@ -0,0 +1,216 @@
...
+		form   *signYourLpaForm
+		errors map[string]string
+	}{
+		"valid": {
+			form: &signYourLpaForm{
+				[]string{"cp-witnessed", "want-to-apply"},
+			},
+			errors: map[string]string{},
+		},
+		"invalid-all": {
```
hawx, line 203: case for selecting one?

app/lang/en.json
```diff
@@ -286,6 +286,19 @@
     "continueToSigningPage": "Continue to signing page",
 
     "signYourLpa": "Sign your LPA",
+    "signYourLpaContent": "<h2 class=\"govuk-heading-m\">Signing the LPA is an important and powerful legal step.</h2> <p class=\"govuk-body\">When you do this you're officially saying that you want {{ .AllAttorneysFullNames }} to be your finance and affairs attorneys.</p> <p class=\"govuk-body\">This means that once the LPA has been registered, {{ .AllAttorneysFirstNames }} will have the legal power to:</p> <ul class=\"govuk-list govuk-list--bullet\"> <li>help you make a decision about your money or property, or make a decision for you if you do not have the mental capacity to make it yourself</li> <li> deal with organisations, like banks and utility companies, on your behalf</li> </ul>",
```
hawx, line 289: I don't think we should be hard-coding "finance and affairs" in here. (And I know I'm not a content person, but what does it mean to be a "finance and affairs attorney"...)

hawx: Thinking about it, we'll need to switch this content depending on how many attorneys too

app/web/template/sign_your_lpa.gohtml
```diff
@@ -6,6 +6,52 @@
...
+
+            {{ template "details" (details . "whatHappensIfIWantToMakeChanges" "whatHappensIfIWantToMakeChangesContent" false) }}
+
+            <hr class="govuk-section-break govuk-section-break--m govuk-section-break--visible">
+
+            {{ trHtml .App "formPanelContent" }}
+
+            <form novalidate method="post">
+                <article class="moj-ticket-panel" aria-label="Sub navigation 1">
+                    <section class="moj-ticket-panel__content moj-ticket-panel__content--blue" aria-label="Section 1">
```
hawx, line 30: These `aria-labels` need removing (or setting to something sensible). But also, if a screen reader user gets a label then why wouldn't someone reading on a screen get one?

hawx: Also these should be `<div>`s rather than `<article>` or `<section>`. It isn't an article, and a section would have a title.

app/web/template/sign_your_lpa.gohtml
```diff
@@ -6,6 +6,52 @@
...
+                        <p class="govuk-body govuk-!-font-weight-bold">
+                            {{ trFormat .App "iConfirmThat" "DonorFullName" .Lpa.DonorFullName }}
+                        </p>
+
+                        {{ template "error-message" (errorMessage . "sign-lpa") }}
+
+                        {{ $wantToApplyRawLabel := trFormat .App "cpIsWitnessingMeSign" "CertificateProviderFullName" .Lpa.CertificateProviderFullName }}
+
+                        {{ template "checkboxes" (items . "sign-lpa" .Form.DonorSignatures
+                        (item "cp-witnessed" "cpIsWitnessingMeSign" "rawLabel" $wantToApplyRawLabel)
```
hawx, line 40: Kind of daft having to pass in `cpIsWitnessingMeSign` when it isn't used. Maybe we should change translation to not error on missing keys, but return it verbatim? That way you wouldn't need `rawLabel`.

app/internal/page/sign_your_lpa.go
acsauk: It was so we could use the `checkbox` template which expects a list of items in order to function correctly (specifically the checked params). Like I said in my PR description, it does feel a bit redundant and I guess I could just use markup rather than relying on the template.

app/internal/page/sign_your_lpa.go
acsauk: This was for protection against the values being unexpected. We've had a few ad hoc convos recently about being able to prove that the user has explicitly agreed to apply and that they were witnessing so checking the values are as they should be felt important here.

app/lang/en.json
acsauk: Ah yeah good point. Will pluralise and pass in the LPA type. 

Also, re: content 🤷

app/web/template/sign_your_lpa.gohtml
acsauk: I missed updating the labeling here from a C&P from MOJ pattern library (also I'm not even sure what we would call this?). That's where the article and section elements were defined - https://design-patterns.service.justice.gov.uk/components/ticket-panel/. I'd want to understand why they went with `article` and `section` in the first place and if changing to `div`s affects any of the UR research on this before changing. I'll see if anyone from MOJ frontend is around still to query with.

app/web/template/sign_your_lpa.gohtml
acsauk: That's a nice workaround. Also just noticed the var naming here is super confusing as its got nothing to do with `wantToApply` 🙄

app/web/template/sign_your_lpa.gohtml
acsauk: Agreed in the context of this declaration element `article` and `content` doesn't make semantic sense. We'll use `div`s instead and drop the `aria-label` as it's not directly linked to an interactive element.

app/internal/page/data.go
```diff
@@ -473,3 +475,23 @@ func (l *Lpa) ReplacementAttorneysStepInWhenOneOrAllAttorneysCannotAct() bool {
...
+}
+
+func (l *Lpa) LpaLegalTerm() string {
+	switch l.Type {
+	case LpaTypePropertyFinance:
+		return "finance and affairs"
+	case LpaTypeHealthWelfare:
+		return "personal welfare"
+	case LpaTypeCombined:
+		return "finance and affairs and personal welfare"
```
hawx, line 494: these will need to be translated

app/web/template/sign_your_lpa.gohtml
```diff
@@ -6,6 +6,72 @@
...
+            <form novalidate method="post">
+                <div class="moj-ticket-panel">
+                    <div class="moj-ticket-panel__content moj-ticket-panel__content--blue">
+                        <p class="govuk-body govuk-!-font-weight-bold">
+                            {{ trFormat .App "iConfirmThat" "DonorFullName" .Lpa.DonorFullName }}
+                        </p>
+
+                        {{ template "error-message" (errorMessage . "sign-lpa") }}
+
+                        {{ $cpWitnessingLabel := trFormat .App "cpIsWitnessingMeSign" "CertificateProviderFullName" .Lpa.CertificateProviderFullName }}
```
hawx, line 47: this is never used

app/lang/en.json
```diff
@@ -23,7 +23,9 @@
     "selectOneOrMoreOptions": "Select one or more options.",
     "answerLater": "Answer later",
     "willDoThisLater": "I will do this later",
-    "pfa": "Finances and affairs",
```
acsauk, line 26: I double checked, we're no longer using this trans key since implementing the humanised version of it

hawx (approved):

PR #251, opened 2022-12-21
MLPAB-509: Witnessing your signature

acsauk:
# Purpose

Adds a guidance page on how the CP should witness donor signing the LPA. Continuing from this page sends an SMS to the CP ready for the next page.

Fixes MLPAB-509, MLPAB-510

## Checklist

* [] The product team have tested these changes

[MLPAB-509]: https://opgtransform.atlassian.net/browse/MLPAB-509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/internal/notify/client.go
```diff
@@ -77,54 +84,90 @@ type errorItem struct {
...
+
+func (c *Client) Sms(ctx context.Context, sms Sms) (string, error) {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(sms); err != nil {
 		return "", err
 	}
-	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("Authorization", "Bearer "+token)
+
+	req, err := c.request(ctx, "/v2/notifications/sms", buf)
```
hawx, line 136: you could probably move all the common code below into this, and call it something like `doRequest`

app/internal/notify/client.go
```diff
@@ -77,54 +84,90 @@ type errorItem struct {
...
-		return "", v.Errors
+	if len(r.Errors) > 0 {
+		return "", r.Errors
 	}
 
-	return v.ID, nil
+	return r.ID, nil
+}
+
+func (c *Client) request(ctx context.Context, url string, jsonBody bytes.Buffer) (*http.Request, error) {
```
hawx, line 156: I'd probably change `bytes.Buffer` to be `io.Reader` so we can be more flexible with what is passed (although we likely never will, so can be ignored)

app/internal/notify/client.go
```diff
@@ -77,54 +84,90 @@ type errorItem struct {
 func (c *Client) TemplateID(name string) string {
 	if c.isProduction {
 		switch name {
-		case "MLPA Beta signature code":
+		case "MLPA Beta signature code - Email":
```
hawx, line 87: I should probably have exported this as a const, but now there are 2 it would make even more sense. Then it could be simplified to an int using `iota` since we don't care about the string value

app/internal/page/certificate_provider_details.go
```diff
@@ -104,6 +108,9 @@ func (d *certificateProviderDetailsForm) Validate() map[string]string {
 	if _, ok := errors["date-of-birth"]; !ok && d.DateOfBirthError != nil {
 		errors["date-of-birth"] = "dateOfBirthMustBeReal"
 	}
+	if d.Mobile == "" {
```
hawx, line 111: should we check it starts with 07 or +447 then has 9 digits following?

acsauk: I mean, it wasn't on the ticket so... 😂 yeah good point, I'll add in.

app/internal/notify/client.go
acsauk: Thanks for this - actually took the time to look into what `io.Reader` actually is and things make a lot more sense in general!

app/internal/notify/client.go
```diff
@@ -107,24 +115,17 @@ func (c *Client) Email(ctx context.Context, email Email) (string, error) {
 		return "", err
 	}
 
-	req, err := c.request(ctx, "/v2/notifications/email", buf)
-
-	resp, err := c.doer.Do(req)
+	req, err := c.request(ctx, "/v2/notifications/email", &buf)
```
hawx, line 118: I was thinking that really all of the method is shared so could be in `doRequest`, but it's fine as is

```
func (c *Client) Email(ctx context.Context, email Email) (string, error) {
  var buf bytes.Buffer
  if err := json.NewEncoder(&buf).Encode(email); err != nil {
    return "", err
  }

  return c.doRequest(ctx, "/v2/notifications/email", &buf)
}
```

hawx (approved):

PR #261, opened 2023-01-09
MLPAB-552 - don't run terraform apply when unit tests fail

andrewpearce-digital:
# Purpose

Deployments ran when unit tests failed, causing a failed deployment because no image could be found.

Fixes MLPAB-552

## Approach

- terraform apply job requires unit test to pass or be skipped

## Checklist


acsauk (approved):
:shipit:

PR #273, opened 2023-01-12
MLPAB-567: Toggle translation keys

acsauk:
# Purpose

Adds a toggle controlled by an env var to display translation keys next to content. This should hopefully make communicating about content changes a little easier as we can refer to the keys that need to be updated rather than relying solely on screenshots and descriptions of the page.

Fixes MLPAB-567

## Approach

I've added the toggle into the footer links for now. If/when we develop further dev features we can work out where to put the menu/links.

## Checklist


app/internal/localize/localizer.go
```diff
@@ -22,32 +23,41 @@ func NewBundle(paths ...string) Bundle {
...
+	return l.translate(l.MustLocalize(&i18n.LocalizeConfig{MessageID: messageID, PluralCount: count}), messageID)
 }
 
 func (l Localizer) FormatCount(messageID string, count int, data map[string]interface{}) string {
 	data["PluralCount"] = count
-	return l.MustLocalize(&i18n.LocalizeConfig{MessageID: messageID, PluralCount: count, TemplateData: data})
+	return l.translate(l.MustLocalize(&i18n.LocalizeConfig{MessageID: messageID, PluralCount: count, TemplateData: data}), messageID)
+}
+
+func (l Localizer) translate(translation, messageID string) string {
```
acsauk, line 57: Had no idea what to call this 🤷

app/web/template/layout/page.gohtml
```diff
@@ -152,6 +152,18 @@
...
                   </a>
                 </li>
+                <li class="govuk-footer__inline-list-item">
+                  <a class="govuk-footer__link" href="#5">
+                    {{ tr .App "termsAndConditions" }}
+                  </a>
+                </li>
+                {{ if eq false .App.IsProduction }}
+                  <li class="govuk-footer__inline-list-item">
+                    <a class="govuk-footer__link" href="{{ if containsStr "showTransKeys=1" .App.Query }}{{ .App.Lang.BuildUrl .App.Page }}{{ else }}{{ .App.Lang.BuildUrl .App.Page }}?showTransKeys=1{{ end }}">
```
hawx, line 162: it would be nicer to have a `.App.ShowTranslationKeys` property instead of using the query

app/web/template/layout/page.gohtml
```diff
@@ -152,6 +152,18 @@
                     {{ tr .App "termsAndConditions" }}
                   </a>
                 </li>
+                <li class="govuk-footer__inline-list-item">
+                  <a class="govuk-footer__link" href="#5">
+                    {{ tr .App "termsAndConditions" }}
+                  </a>
+                </li>
+                {{ if eq false .App.IsProduction }}
```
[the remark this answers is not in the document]
acsauk: Ah cool - still getting used to conditionals in go templates

.github/workflows/workflow_path_to_live.yml
```diff
@@ -110,13 +110,23 @@ jobs:
       version_tag: ${{ needs.create_tags.outputs.version_tag }}
     secrets: inherit
 
+  production_smoke_tests:
```
hawx, line 113: I've been thinking about this, maybe we should just call them "production tests" (so under `cypress/production`), since that is what they are for?

app/internal/templatefn/fn.go
```diff
@@ -132,6 +134,10 @@ func contains(needle string, list []string) bool {
 	return slices.Contains(list, needle)
 }
 
+func containsStr(needle, s string) bool {
```
hawx, line 137: you don't need this now

app/internal/page/app.go
```diff
@@ -458,17 +461,21 @@ func makeHandle(mux *http.ServeMux, logger Logger, store sessions.Store, localiz
...
+				Localizer:           localizer,
+				Lang:                lang,
+				SessionID:           sessionID,
+				CookieConsentSet:    cookieErr != http.ErrNoCookie,
+				CanGoBack:           opt&CanGoBack != 0,
+				RumConfig:           rumConfig,
+				StaticHash:          staticHash,
+				Paths:               paths,
+				IsProduction:        isProduction,
+				ShowTranslationKeys: r.FormValue("showTranslationKeys") == "1",
```
hawx, line 478: I didn't think when I commented before, you have `Localizer` in here so could just use `.App.Localizer.ShowTranslationKeys`

.github/workflows/workflow_path_to_live.yml
acsauk: Yeah was chatting this through with @andrewpearce-digital earlier - we're going to add in another check before preproduction cypress tests run to catch any misconfigurations before we deploy to prod so would agree the name change makes sense.

I'll still leave the prod test as well in case the tf config for preproduction drifts out of sync over time.

andrewpearce-digital (approved):

terraform/environment/variables.tf
```diff
@@ -27,6 +27,7 @@ variable "environments" {
           yoti_client_sdk_id     = string
           yoti_scenario_id       = string
           yoti_sandbox           = string
+          dev_features_enabled   = string
```
hawx, line 30: if it's enabled everywhere could it not just be deleted?

PR #275, opened 2023-01-16
MLPAB-529 Use lpa.Tasks to keep track of all tasks

hawx:
The pros of doing this vs. calculating are that it stops us duplicating logic to check things and when one day we redo the data "properly" we won't have to grab the full LPA each time to render the task list.


acsauk (approved):
So much simpler - nice job :shipit:

PR #280, opened 2023-01-18
MLPAB-554: Handle postcode errors

acsauk:
# Purpose

Adds some error handling when users enter a postcode that is not found via the OS API and enforces a value to entered in the postcode field of the manual entry form.

Now we have standardised validations and a base address form I've moved our e2e assertions into some reusable functions rather than having to update 5 separate files every time we change some address form content or behaviour. 

Fixes MLPAB-554.

## Checklist


cypress/e2e/people-to-notify.cy.js
```diff
@@ -1,3 +1,5 @@
+import {AddressFormAssertions} from "../support/e2e";
```
acsauk, line 1: I've left the format of this test relying on fixture objects intact for now until we tackle the fixtures ticket as we're still undecided on how this will work in practice.

app/internal/page/address_form.go
```diff
@@ -63,6 +63,8 @@ func (f *addressForm) Validate() validation.List {
 			validation.StringTooLong(50))
 		errors.String("address-town", "townOrCity", f.Address.TownOrCity,
 			validation.Empty())
+		errors.String("address-postcode", "aPostcode", f.Address.Postcode,
+			validation.Empty())
```
hawx, line 67: I think the reason this was originally left out was because we need to allow international addresses (although I imagine that may need other changes too, so happy for it be added in on this I guess)

app/internal/page/certificate_provider_address.go
```diff
@@ -64,7 +65,13 @@ func CertificateProviderAddress(logger Logger, tmpl template.Template, addressCl
 				addresses, err := addressClient.LookupPostcode(r.Context(), data.Form.LookupPostcode)
 				if err != nil {
 					logger.Print(err)
-					data.Errors.Add("lookup-postcode", validation.CustomError{Label: "couldNotLookupPostcode"})
+
+					if errors.As(err, &place.InvalidPostcodeError{}) {
```
hawx, line 69: since you aren't using the unwrapped error these could probably use `errors.Is` instead

app/internal/page/certificate_provider_address.go
```diff
@@ -64,7 +65,13 @@ func CertificateProviderAddress(logger Logger, tmpl template.Template, addressCl
...
+						data.Errors.Add("lookup-postcode", validation.EnterError{Label: "invalidPostcode"})
```
hawx, line 70: On https://design-system.service.gov.uk/patterns/addresses/ they use "Enter a real postcode" which might be nicer than the word "valid"?

app/internal/page/address_form.go
```diff
@@ -46,11 +46,11 @@ func (f *addressForm) Validate() validation.List {
 
 	switch f.Action {
 	case "lookup":
-		errors.String("lookup-postcode", "postcode", f.LookupPostcode,
+		errors.String("lookup-postcode", "aPostcode", f.LookupPostcode,
```
hawx, line 49: are we going to want a ticket to add "a/an" for all field error messages where it makes sense?

app/internal/place/client.go
```diff
@@ -78,6 +89,10 @@ func (c *Client) LookupPostcode(ctx context.Context, postcode string) ([]Address
 		return []Address{}, err
 	}
 
+	if postcodeLookupResponse.Error.Statuscode == invalidPostcodeCode {
```
hawx, line 92: `invalidPostcodeCode` is just `http.StatusBadRequest`, so also should the error be `BadRequestError` since it isn't necessarily an invalid postcode?

app/internal/place/client.go
```diff
@@ -33,13 +34,23 @@ type addressDetails struct {
...
+	Error   InvalidPostcodeError `json:"error"`
 }
 
 type ResultSet struct {
 	AddressDetails addressDetails `json:"DPA"`
 }
 
+type InvalidPostcodeError struct {
+	Statuscode int    `json:"statuscode,omitempty"`
+	Message    string `json:"message,omitempty"`
```
hawx, line 47: `omitempty` is only relevant for things we serialise to JSON

app/lang/en.json
```diff
@@ -87,7 +91,7 @@
     "whichTypeOfLpaToMake": "Which type of LPA do you want to make?",
     "whichTypeOfLpaToMakeContent": "<p class=\"govuk-body\">There are 2 types of LPA. You can make an LPA that covers decisions about your:</p><ul class=\"govuk-list govuk-list--bullet\"><li>money, finances and any property you own</li><li>health care, medical treatment and personal welfare</li></ul><p class=\"govuk-body\">You can have one or both types of LPA.</p><p class=\"govuk-body\">They do not have to be made at the same time and you can reuse any information you enter to make other LPAs later.</p><p class=\"govuk-body\">You can choose different people to be your attorneys for each LPA.</p><p class=\"govuk-body\"><a class=\"govuk-link\" href=\"#\">Find out more about the 2 types of LPAs</a></p>",
     "theTypeOfLpaToMake": "the type of LPA to make",
-    "lpaTypePfa": "Property and affairs",
+    "lpaTypePfa": "Finance and affairs",
```
hawx, line 94: https://opgtransform.atlassian.net/browse/MLPAB-555 is this getting changed again?

app/internal/page/address_form.go
acsauk: Yeah, worth running through with Amy and Sophie on recent validation error changes so we can hit them up in one go.

app/internal/page/certificate_provider_address.go
SophieDodds: Good question but we've deliberately chosen to go with **valid** over **real** based on conversations around this in the content community. ONS and other design systems also use valid over real: 

From ONS:  If the postcode entered is not valid
Use “Enter a valid UK postcode”. 

See https://ons-design-system.netlify.app/patterns/addresses/#manual-addresses 

"Real" to the user is not helpful if they are entering a postcode they believe to be real. I think it sounds like we are casting doubt on the user's input. Whereas 'valid' seems a bit more neutral. Other teams have also had this discussion and ended up going with 'valid'. Can send Slack threads about this, if interested.

app/internal/place/client.go
acsauk: That's true - it's a shame the API doesn't allow us to differentiate any further than this. I'll update the error type name so its more generic.

app/internal/place/client.go
acsauk: Agh, I always get this the wrong way round - thanks, will update.

app/lang/en.json
acsauk: So I got really confused during bringing `main` in here and had convinced myself `Finance and affairs` was the new term. I'll revert!

app/internal/page/certificate_provider_address.go
acsauk: I tried this initially (and have just tried again now in case I got `target` and `err` the wrong way round) but `Is()` doesn't seem to recognise `err` as a `place.InvalidPostcodeError`. 

The whole piece on unwrapping errors confused me - is this to try to get down to a base type somehow?

hawx: You'd need to change the error from being a struct to just a value to use `Is`, which would be fine since you aren't using the Message or Statuscode. But yeah, more change so don't know if you care about doing it

app/internal/page/certificate_provider_address.go
```diff
@@ -64,7 +65,16 @@
...
 					logger.Print(err)
-					data.Errors.Add("lookup-postcode", validation.CustomError{Label: "couldNotLookupPostcode"})
+
+					if errors.As(err, &place.BadRequestError{}) {
+						data.Errors.Add("lookup-postcode", validation.EnterError{Label: "invalidPostcode"})
+					} else {
+						data.Errors.Add("lookup-postcode", validation.CustomError{Label: "couldNotLookupPostcode"})
+					}
+				} else if len(addresses) == 0 {
+					logger.Print("No addresses found for " + data.Form.LookupPostcode)
```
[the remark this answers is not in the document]
hawx: I'd say we don't need these logs since we won't take any action from them

app/internal/place/client.go
hawx: yep, I'd love them to return a better body than just "message"

app/internal/page/certificate_provider_address.go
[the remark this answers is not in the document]
acsauk: @hawx I've wondered about how our logging works for a while - is there any kind of auto-escaping/sanitizing going on behind the scene in the logging package we use or will this be raw input?

In this case we're pretty sure its at least a valid form postcode otherwise we'd get a 400 response but still feels like we should be cautious with user input in logs.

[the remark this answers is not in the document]
acsauk: Good point - logging probably isn't the best place for this but I'd still be interested in how often this happens. Maybe something to look at with RUM if we decide to track it.

hawx (approved):

PR #283, opened 2023-01-19
MLPAB-412: Fix people to notify summary redirect and completed status not being set

acsauk:
# Purpose

Small fix I noticed when testing some content changes out. When we get to 5 people to notify the form changes to a link (which was going to the wrong page). I've replaced this with a form and hidden input so we rely on the handler logic to set the redirect URL. Status updates have also been aligned with how adding attorneys work (e.g. set to in progress on adding initial details and set to complete on entering address)

Also fixes up some content on the initial people to notify page.

Fixes MLPAB-412.


SophieDodds (approved):
Discussed error summary order - to be addressed in another ticket - this is fine and ready to be merged.

PR #284, opened 2023-01-19
MLPAB-555 Change term for pfa to property and affairs

hawx:


acsauk (approved):
On various vars, translation keys and other strings we use `pfa` and `hw` (and other expanded variations) which could confuse future maintainers if they don't correspond to the terms we use in content (e.g. `pfa` -> `pa` and `hw` -> `pw`). It may be worth having a chat through if it's worth aligning with our content or staying aligned with how they are referred to in the wider estate.

PR #286, opened 2023-01-20
MLPAB-306 Move LPA specific handlers under /lpa/ route

hawx:
This is just the refactoring, the actual change for multiple LPAs will be done in a separate PR to stop this becoming even larger.

### DIFF
--- a/app/internal/page/app.go
+++ b/app/internal/page/app.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
@@ -30,23 +31,6 @@ type RumConfig struct {
 
 type Lang int
 
-func CacheControlHeaders(h http.Handler) http.Handler {
-	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		w.Header().Set("Cache-Control", "max-age=2592000")
-		h.ServeHTTP(w, r)
-	})
-}
-
-func (l Lang) Redirect(w http.ResponseWriter, r *http.Request, lpa *Lpa, url string) error {
-	// as a shortcut for when you don't have an Lpa but know the transition is fine we allow passing nil
-	if lpa == nil || lpa.CanGoTo(url) {
-		http.Redirect(w, r, l.BuildUrl(url), http.StatusFound)
-	} else {
-		http.Redirect(w, r, l.BuildUrl(Paths.TaskList), http.StatusFound)
-	}
-	return nil
-}
-
 func (l Lang) String() string {
 	if l == Cy {
 		return welshAbbreviation
@@ -55,12 +39,17 @@ func (l Lang) String() string {
 	return englishAbbreviation
 }
 
-func (l Lang) BuildUrl(url string) string {
-	if l == Cy {
-		return "/" + welshAbbreviation + url
-	} else {
-		return url
-	}
+func CacheControlHeaders(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Cache-Control", "max-age=2592000")
+		h.ServeHTTP(w, r)
+	})
+}
+
+func IsLpaPath(url string) bool {
+	path, _, _ := strings.Cut(url, "?")
+
+	return path != Paths.Dashboard
 }
 
 const (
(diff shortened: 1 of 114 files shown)


app/internal/page/app.go
```diff
@@ -55,12 +39,17 @@ func (l Lang) String() string {
...
+		w.Header().Set("Cache-Control", "max-age=2592000")
+		h.ServeHTTP(w, r)
+	})
+}
+
+func IsLpaPath(url string) bool {
+	path, _, _ := strings.Cut(url, "?")
+
+	return path != Paths.Dashboard
 }
```
hawx, line 53: This isn't great, but it does solve the problem for now. I'm thinking eventually it would be nice to make the path definitions as structs so we can attach a bit more info to them, and make them self formatting (like `Paths.YourDetails.Format(123) == "/lpa/123/your-details"` maybe)

app/internal/page/data.go
```diff
@@ -161,42 +161,80 @@ func readDate(t time.Time) Date {
...
 
-	if lpa.ID == "" {
-		lpa.ID = "10" + strconv.Itoa(s.randomInt(100000))
+	return &lpa, err
+}
+
+func (s *lpaStore) Get(ctx context.Context) (*Lpa, error) {
+	data := sessionDataFromContext(ctx)
+	if data.LpaID == "" {
+		return nil, fmt.Errorf("lpaStore.Get requires LpaID to retrieve")
```
acsauk, line 225:
```suggestion
		return nil, fmt.Error("lpaStore.Get requires LpaID to retrieve")
```

app/internal/page/data.go
```diff
@@ -161,42 +161,80 @@ func readDate(t time.Time) Date {
...
+
+func (s *lpaStore) Create(ctx context.Context) (*Lpa, error) {
+	lpa := &Lpa{ID: "10" + strconv.Itoa(s.randomInt(100000))}
+	err := s.Put(ctx, lpa)
+
+	return lpa, err
 }
 
-func (s *lpaStore) Get(ctx context.Context, sessionID string) (*Lpa, error) {
+func (s *lpaStore) GetAll(ctx context.Context) (*Lpa, error) {
```
acsauk, line 215: I'm assuming this will return `[]*Lpa` on the next ticket?

app/internal/page/app.go
acsauk: That sounds like a good idea and this is totes fine for now 👍

app/internal/page/app.go
```diff
@@ -138,120 +159,126 @@ func App(
...
 		Dashboard(tmpls.Get("dashboard.gohtml"), lpaStore))
-	handle(paths.LpaType, RequireSession,
-		LpaType(tmpls.Get("lpa_type.gohtml"), lpaStore))
-	handle(paths.WhoIsTheLpaFor, RequireSession,
-		WhoIsTheLpaFor(tmpls.Get("who_is_the_lpa_for.gohtml"), lpaStore))
 
-	handle(paths.YourDetails, RequireSession,
+	lpaMux := http.NewServeMux()
+
+	rootMux.Handle("/lpa/", routeToLpa(lpaMux))
```
acsauk, line 179: I was wondering how path scoping worked in Go, nice!

app/internal/page/app.go
```diff
@@ -481,6 +524,29 @@ func makeHandle(mux *http.ServeMux, logger Logger, store sessions.Store, localiz
...
+			return
+		}
+
+		id, path := parts[2], "/"+parts[3]
+
+		r2 := new(http.Request)
+		*r2 = *r
+		r2.URL = new(url.URL)
+		*r2.URL = *r.URL
+		r2.URL.Path = path
```
acsauk, line 541: OK I'm lost...is this how things are cloned in Go?

app/internal/page/app.go
```diff
@@ -481,6 +524,29 @@ func makeHandle(mux *http.ServeMux, logger Logger, store sessions.Store, localiz
...
+		if len(r.URL.RawPath) > 5+len(id) {
```
acsauk, line 542: Is the significance of `5` here `/lpa/`?

app/internal/page/data.go
```diff
@@ -161,42 +161,80 @@ func readDate(t time.Time) Date {
...
+func (d *sessionData) String() string {
+	if d == nil {
+		return "sessionData<nil>"
+	}
+
+	return fmt.Sprintf("sessionData(sessionID=%s lpaID=%s)", d.SessionID, d.LpaID)
+}
+
+func sessionDataFromContext(ctx context.Context) *sessionData {
+	data, _ := ctx.Value((*sessionData)(nil)).(*sessionData)
```
acsauk, line 199: Whats the purpose of the `(nil)` here?

acsauk (commented):
Mega job! I've got a few Qs on some specific bits of code and one suggestion but otherwise good to go.

app/internal/page/data.go
hawx: yep, that is the plan

app/internal/page/app.go
hawx: yeah, this is pretty much the code for [`http.StripPrefix`](https://cs.opensource.google/go/go/+/refs/tags/go1.19.5:src/net/http/server.go;l=2138), but modified to deal with the ID in the URL

app/internal/page/app.go
hawx: yes, I will constantify it

app/internal/page/data.go
hawx: I'm creating a `nil` of type `*sessionData`. So I get a unique key for the context, using nil which means it doesn't allocate. It's a one line way of doing

```
var key *sessionData
data, _ := ctx.Value(key).(*sessionData)
```

acsauk: Right, making sense now. It hadn't clicked that `nil` was being used as the value. So would this also be valid (not that it would make sense for what we need here):

```go
data, _ := ctx.Value(&sessionData{someProp: 'someValue'}).(*sessionData)
```

Happy to keep it as is now I can see the purpose.

hawx: that would be valid if you'd set something with the same key

PR #289, opened 2023-01-23
MLPAB-306 Allow creating multiple LPAs

hawx:


app/internal/dynamo/client.go
```diff
@@ -23,17 +24,39 @@ func NewClient(cfg aws.Config, tableName string) (*Client, error) {
...
+	pkey, err := attributevalue.Marshal(pk)
+	if err != nil {
+		return err
+	}
+
+	response, err := c.svc.Query(ctx, &dynamodb.QueryInput{
+		TableName:                 aws.String(c.table),
+		ExpressionAttributeNames:  map[string]string{"#0": "PK"},
+		ExpressionAttributeValues: map[string]types.AttributeValue{":0": pkey},
+		KeyConditionExpression:    aws.String("#0 = :0"),
```
hawx, line 37: there is an expression builder to construct these, but :shrug: it seemed about as clear what was happening, and at least this matches the docs

terraform/environment/dynamodb.tf
```diff
@@ -9,13 +9,13 @@ data "aws_kms_alias" "dynamodb_encryption_key_eu_west_2" {
 }
 
 resource "aws_dynamodb_table" "lpas_table" {
-  name         = "${local.environment_name}-Lpas"
+  name         = "${local.environment_name}-Lpas2"
```
hawx, line 12: need this to be recreated, and the actual name doesn't really matter so... added a 2 😄

app/internal/dynamo/client.go
acsauk: If I understand this right, we've set the value of `#0` to `"PK"` (Primary Key?) in `ExpressionAttributeNames` and the value of `:0` to `pkey` and we're then just setting a straight equality check in `KeyConditionExpression`? Could `#0` and `:0` be any string? It seemed that way from the docs but wanted to check as I've not worked with dynamoDB before.

localstack-init.sh
```diff
@@ -10,6 +10,6 @@ awslocal secretsmanager create-secret --name "os-postcode-lookup-api-key" --secr
 awslocal secretsmanager create-secret --name "yoti-private-key" --secret-string "bm90aGluZwo="
 awslocal secretsmanager create-secret --name "gov-uk-notify-api-key" --secret-string "extremely_fake-a-b-c-d-e-f-g-h-i-j"
 
-awslocal dynamodb create-table --table-name lpas --attribute-definitions AttributeName=Id,AttributeType=S --key-schema AttributeName=Id,KeyType=HASH --provisioned-throughput ReadCapacityUnits=1000,WriteCapacityUnits=1000
+awslocal dynamodb create-table --table-name lpas --attribute-definitions AttributeName=PK,AttributeType=S AttributeName=SK,AttributeType=S --key-schema AttributeName=PK,KeyType=HASH AttributeName=SK,KeyType=RANGE --provisioned-throughput ReadCapacityUnits=1000,WriteCapacityUnits=1000
```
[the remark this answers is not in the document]
acsauk: Cool - I'd drafted a comment above and then scrapped it after reading this. Makes sense 👍

terraform/environment/dynamodb.tf
acsauk: Is the issue here that terraform won't recreate the resource when changing the hash key?

app/internal/dynamo/client.go
hawx: I guess I should not have followed whatever thing I saw verbatim and called them `#PK` and `:PK` instead

terraform/environment/dynamodb.tf
hawx: yeah you can't change the keys on an existing table

app/internal/dynamo/client.go
acsauk: 😁 may be slightly clearer

acsauk (approved):
:shipit:

terraform/environment/dynamodb.tf
andrewpearce-digital: That would have resulted in a recreation anyway

PR #295, opened 2023-01-27
MLPAB-589 Add different <title>s for each page

hawx:
https://github.com/alphagov/govuk-design-system-backlog/issues/157 has the same advice of personalise `<h1>` but not `<title>` that we were thinking, but weird that there isn't an actual page on GDS of "the" way to format them beyond [this](https://www.gov.uk/service-manual/design/writing-for-user-interfaces#style)


acsauk (approved):
Seems a good approach as any given lack of GDS guidance 🤷 

:shipit:

PR #296, opened 2023-01-27
MLPAB-590 Use a list for validation errors so they order correctly

hawx:
The main thing here is the new `validation` package and `List` type, which will help us centralise more of this logic. Everything else is me replacing the old `map[string]string`. I'm not 100% on the `.With` naming, but it will at least be easy to rename in the future if there is a better name.


acsauk (approved):
This has really improved template and code readability 🙌 

:shipit:

PR #301, opened 2023-01-31
MLPAB-636: Conditionally show how attorneys work together

acsauk:
# Purpose

Ensures we don't end up with an empty bullet on `/do-you-want-to-notify-people` when a single attorney is set.

Fixes MLPAB-636


amywilson-8 (dismissed):
Bullet only appears for multiple deputies and grammar adapts correctly (your attorneys are/your attorney is). Meets ticket requirements

hawx (approved):

PR #303, opened 2023-01-31
MLPAB-563 Add validation package and use templates for error messages

hawx:


app/internal/date/date.go
```diff
@@ -0,0 +1,32 @@
+package date
```
hawx, line 1: I want to improve this in a later PR. I'm thinking it should have Un/MarshalText methods so we can use it in structs where we want a date, and not a time. Also promote `After`/`Before` from `T` so we can hide the underlying `time.Time` value.

app/internal/page/certificate_provider_address.go
```diff
@@ -117,36 +117,28 @@ func readCertificateProviderAddressForm(r *http.Request) *certificateProviderAdd
 	return d
 }
 
-func (d *certificateProviderAddressForm) Validate() validation.List {
+func (f *certificateProviderAddressForm) Validate() validation.List {
```
hawx, line 120: I'm 99% sure all the address forms are the same, so want to consolidate them (in a separate PR)

app/internal/page/check_your_lpa.go
```diff
@@ -70,13 +68,10 @@ func readCheckYourLpaForm(r *http.Request) *checkYourLpaForm {
...
-		errors.Add("checked", "selectCheckedLpa")
-	}
-
-	if !f.Happy {
-		errors.Add("happy", "selectHappyToShareLpa")
-	}
+	errors.Bool("checked", "checkedLpa", f.Checked,
+		validation.Selected())
+	errors.Bool("happy", "happyToShareLpa", f.Happy,
+		validation.Selected())
```
hawx, line 74: Some of these are kind of only relevant with one type of check, so again, in a future PR I want to tidy up the API  to make more sense

app/lang/en.json
```diff
@@ -35,22 +35,27 @@
...
-    "firstNamesTooLong": "Enter a first name that's less than 54 characters long",
-    "enterLastName": "Enter last name",
-    "lastNameTooLong": "Enter a last name that's less than 62 characters long",
-    "otherNamesTooLong": "Enter other names that are less than 51 characters long",
-    "enterDateOfBirth": "Enter date of birth",
-    "dateOfBirthMustBeReal": "Date of birth must be a real date",
-    "dateOfBirthIsFuture": "Date of birth must be before today",
+    "day": "day",
+    "month": "month",
+    "year": "year",
```
acsauk, line 43: I get why these have been lowercased but when we rely on them for form labels we're getting:

Screenshot 2023-02-01 at 13 26 44

I guess we just need to capitalise in `date.gohtml`?

app/internal/page/certificate_provider_details.go
```diff
@@ -88,25 +83,20 @@ var mobileRegex = regexp.MustCompile(`^(?:07|\+?447)\d{9}$`)
 func (d *certificateProviderDetailsForm) Validate() validation.List {
```
acsauk, line 83: Can't comment on the actual line as its not changed but I think we can drop `mobileRegex` now its centralised in validation.

acsauk (changes requested):
I started to make comments on the func naming in the validation package and then deleted them once I saw how it was used - super elegant 👌 

Couple of comments to action but looks good to go other than that.

app/lang/en.json
hawx: good spot

acsauk (approved):
:shipit: 

May need to run some of the validation error changes past Amy and Sophie as I think they're keeping a spreadsheet with a list of them to ensure they're consistent. Once we have weblate I'm sure there will be a better way to keep on top of it all.

PR #310, opened 2023-02-07
MLPAB-376 Make cache refresh secrets after a specified duration

hawx:
This is potentially still over the top for what we have. I figured there was no point finding an LRU cache to match what we had before because the default size of 1024 was way over what we'd ever need. I've matched the 1 hour TTL but not gone as far as to implement exponential backoff (just doing a simple `errors * delay` instead).

### DIFF
--- a/app/internal/secrets/secrets.go
+++ b/app/internal/secrets/secrets.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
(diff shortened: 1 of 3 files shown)


app/internal/secrets/secrets.go
```diff
@@ -20,42 +21,75 @@ const (
...
 	}
 
 	result, err := c.svc.GetSecretValue(ctx, &secretsmanager.GetSecretValueInput{
 		SecretId: aws.String(name),
 	})
 	if err != nil {
+		if found {
+			if item.errorCount < maxErrorCount {
+				item.errorCount++
+			}
```
acsauk, line 77: Are we accounting for potential network issues with `maxErrorCount`?

hawx: `maxErrorCount` was just to limit the delay growth, but happy to remove since I think it is unlikely to be a problem

acsauk (approved):
:shipit:

PR #311, opened 2023-02-08
MLPAB-328 Warn when actors have the same names

hawx:


app/internal/actor/type.go
```diff
@@ -0,0 +1,97 @@
+package actor
```
hawx, line 1: There is some stuff that we could move out of `page/data.go` into this package, so will take a look after this PR.

acsauk (approved):
:shipit:

PR #312, opened 2023-02-09
MLPAB-328 Move actor related stuff out of data.go into actor

hawx:
One day I will stop changing every file 😅


acsauk (approved):
I think you secretly love it :shipit:

PR #313, opened 2023-02-09
MLPAB-660 Start page and ID for certificate provider

hawx:
I've also added https://opgtransform.atlassian.net/browse/MLPAB-669 and https://opgtransform.atlassian.net/browse/MLPAB-670 to the backlog because I think the data should be done differently.

### DIFF
--- a/app/internal/page/app.go
+++ b/app/internal/page/app.go
@@ -10,16 +10,15 @@ import (
 	"strings"
 	"time"
 
-	"github.com/ministryofjustice/opg-modernising-lpa/internal/actor"
-	"github.com/ministryofjustice/opg-modernising-lpa/internal/onelogin"
-	"github.com/ministryofjustice/opg-modernising-lpa/internal/place"
-
 	"github.com/gorilla/sessions"
 	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/actor"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/identity"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/localize"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/notify"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/onelogin"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/pay"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/place"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/random"
 )
 
@@ -46,12 +45,6 @@ func CacheControlHeaders(h http.Handler) http.Handler {
 		w.Header().Set("Cache-Control", "max-age=2592000")
 		h.ServeHTTP(w, r)
 	})
-}
-
-func IsLpaPath(url string) bool {
-	path, _, _ := strings.Cut(url, "?")
-
-	return path != Paths.Dashboard && path != Paths.Start
 }
 
 const (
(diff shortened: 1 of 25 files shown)


app/internal/page/app.go
```diff
@@ -175,16 +169,30 @@ func App(
...
 		Dashboard(tmpls.Get("dashboard.gohtml"), lpaStore))
 
+	handleRoot(paths.CertificateProviderStart, None,
+		CertificateProviderStart(tmpls.Get("certificate_provider_start.gohtml"), lpaStore))
+	handleRoot(paths.CertificateProviderLogin, None,
+		CertificateProviderLogin(logger, oneLoginClient, sessionStore, random.String))
+	handleRoot(paths.CertificateProviderLoginCallback, None,
+		CertificateProviderLoginCallback(tmpls.Get("identity_with_one_login_callback.gohtml"), oneLoginClient, sessionStore, lpaStore))
+	handleRoot(paths.CertificateProviderYourDetails, RequireCertificateProviderSession,
+		Guidance(tmpls.Get("certificate_provider_your_details.gohtml"), "", lpaStore))
```
hawx, line 182: I want to move the certificate provider stuff to it's own package, and rename `page`, but not sure what to call them at the moment other than `certifypage` (I think `certificateproviderpage` would be crazy, but maybe?) and `donorpage`

app/internal/page/app.go
```diff
@@ -535,6 +549,32 @@ func makeHandle(mux *http.ServeMux, logger Logger, store sessions.Store, localiz
 				}
 			}
 
+			if opt&RequireCertificateProviderSession != 0 {
```
hawx, line 552: or do `RequireSession` and `CertificateProvider` so you'd set both (otherwise I need to put a check in to prevent both being set)

app/internal/page/app.go
```diff
@@ -515,14 +524,19 @@ func makeHandle(mux *http.ServeMux, logger Logger, store sessions.Store, localiz
 					return
 				}
 
-				sub, ok := session.Values["sub"].(string)
+				params, ok := session.Values["donor"].(*DonorLoginSession)
```
hawx, line 527: I'm considering writing a helper for this, so we'd have `donorLoginSession, ok := getDonorSession(store)`, etc.

hawx: and could probably return `nil` instead of `, ok`

app/internal/page/auth_redirect.go
```diff
@@ -1,12 +1,56 @@
...
+	State               string
+	Nonce               string
+	Locale              string
+	Identity            bool
+	CertificateProvider bool
+	SessionID           string
+	LpaID               string
+}
+
+func (s OneLoginSession) Valid() bool {
```
hawx, line 20: these could return `error` instead?

app/internal/page/certificate_provider_start.go
```diff
@@ -0,0 +1,44 @@
...
+
+		lpa, err := lpaStore.Get(contextWithSessionData(r.Context(), &sessionData{
+			SessionID: sessionID,
+			LpaID:     lpaID,
+		}))
+		if err != nil {
+			return err
+		}
+		if lpa == nil {
+			return fmt.Errorf("lpa does not exist")
```
hawx, line 34: if this doesn't exist maybe just show a more generic page?

app/internal/page/app.go
```diff
@@ -175,16 +168,29 @@ func App(
...
+	handleRoot(paths.CertificateProviderYourDetails, RequireSession|RequireCertificateProvider,
+		Guidance(tmpls.Get("certificate_provider_your_details.gohtml"), "", lpaStore))
+
 	lpaMux := http.NewServeMux()
 
 	rootMux.Handle("/lpa/", routeToLpa(lpaMux))
 
 	handleLpa := makeHandle(lpaMux, logger, sessionStore, localizer, lang, rumConfig, staticHash, paths, RequireSession)
 
+	handleLpa("/testing-certificate-provider-start", None, func(appData AppData, w http.ResponseWriter, r *http.Request) error {
```
hawx, line 189: this is just temporary so we have something to go to before getting to sending the email

app/internal/page/app.go
acsauk: In keeping with the action focus of`donorpage` maybe `certifierpage`? Seeing it written down I don't like it much.   although I gues technically following that logic `donorpage` would actually be `donatorpage` which is also not great 🤔.

`certificateproviderpage` is long but if we want to follow the actor+page pattern it fits 🤷

app/internal/page/sessions.go
```diff
@@ -0,0 +1,170 @@
...
+	return fmt.Sprintf("%s session invalid", string(e))
+}
+
+var (
+	sessionCookieOptions = &sessions.Options{
+		Path:     "/",
+		MaxAge:   24 * 60 * 60,
+		SameSite: http.SameSiteLaxMode,
+		HttpOnly: true,
+		Secure:   true, // check this works
```
acsauk, line 29: Did it work? 🤔

acsauk (commented):
Couple of suggestions/Qs and the ready to merge

app/internal/page/sessions.go
hawx: ha, yes, and this is why I should put TODO before my comments so I actually spot them

app/web/template/layout/page.gohtml
```diff
@@ -157,6 +157,11 @@
                     {{ tr .App "toggleTransKeys" }}
                   </a>
                 </li>
+                <li class="govuk-footer__inline-list-item">
+                  <a class="govuk-footer__link" href="testing-certificate-provider-start">
+                    Certificate provider start
```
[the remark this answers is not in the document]
hawx: this should be deleted when we get the email done, I don't like it

app/internal/page/sessions.go
```diff
@@ -0,0 +1,170 @@
...
+	session.Values = map[any]any{"donor": donorSession}
+	session.Options = sessionCookieOptions
+	return store.Save(r, w, session)
+}
+
+type CertificateProviderSession struct {
+	Sub       string
+	Email     string
+	LpaID     string
+	SessionID string // this is the donor's sessionID
```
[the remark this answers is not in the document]
hawx: :+1: smart

app/web/template/layout/page.gohtml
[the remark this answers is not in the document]
acsauk: I still have a strong desire for some kind of admin view somewhere that we can hide these things in (including the translations toggle). I've usually relied on user roles for this but as we scope the roles to types rather than prop on a user type we'd need a different approach for modernise...

acsauk (approved):
:shipit:

PR #315, opened 2023-02-10
MLPAB-594: Add CSRF protection to all forms

acsauk:
# Purpose

Adds double-submit CSRF protection to all forms in the app.

Fixes MLPAB-594

## Approach

I've made a few attempts at making this as straightforward as possible and ended up here. Ideally I'd have liked to wrap the form element with the csrf hidden field in a template partial but I was coming up against what looks to be a limit of using `{{ define }}` inside another define block e.g.:

```html
{{/*  page with form */}}
{{ define "main" }}
  <div class="govuk-grid-row">
      {{ template "form" }}
        {{ define "form-content" }}
          ... form markup here
        {{ end }}
      {{ end }}
    </div>
  </div>
{{ end }}

{{/*  form template */}}
{{ define "form" }}
  <form novalidate method="post">
    {{ template "form-content" . }}
    <input type="hidden" name="csrf" value="{{ .App.CsrfToken }}" />
  </form>
{{ end }}
```

There was also an annoying cypress bug when directly visiting a URL using `visitLpa()` as opposed to navigation to a page where the csrf token from the previous request was persisting. A page reload resyncs to csrf values so I've added that option to the command but it doesn't feel great 😞.

## Learning

Great [cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie) on handling CSRF from OWASP. I've exclusively relied on frameworks to handle this for me before so was good to wrap my head around the options and vulnerability itself.

## Checklist


app/internal/page/app.go
```diff
@@ -489,11 +490,28 @@ const (
...
+					return
+				}
+
+				if !csrfValid(r, csrfSession) {
+					http.Error(w, "CSRF token not valid", http.StatusForbidden)
+					return
+				}
+			}
+
+			csrfSession.Values = map[interface{}]interface{}{"token": randomString(12)}
```
hawx, line 512: I think we should only set this if not already set, that way you'll still be able to press the back button and submit a form again.

app/internal/page/app.go
```diff
@@ -489,11 +490,28 @@ const (
...
+			_ = store.Save(r, w, csrfSession)
```
hawx, line 513: also we'll want to set `HttpOnly`, `Secure`, `SameSiteStrict`(?) and a sensible expiry (a day?) on the cookie

app/internal/page/app.go
acsauk: Yeah, I had thought about setting this once per period of time/session but I wasn't sure if this made things less secure. Happy to keep it around for a day and this will hopefully fix the cypress visitLpaPath weirdness I was seeing.

hawx (approved):

PR #322, opened 2023-02-13
MLPAB-660 Split page package by actor

hawx:
maybe I do love changing all the files

I made an `app` package to prevent an import cycle in `page`, it's kind of lame but solves the problem. This has put some things in weird places, like `lpaStore`, that can be solved later when we expand the data model. Also some stuff (mainly for tests) is now duplicated, but this could be cut down in the future. 

I've moved the session helpers to a `sesh` package so we can start to isolate this stuff before deciding what to do with `gorilla/sessions`. I can change the name if you hate, I just didn't want to call it `sessions` and then have to figure out which import URL is needed.

I'm leaving `paths.go` in `page` for now, but it would probably make sense to split this up at some point. Then we can prevent ourselves linking between the 2 journeys, or at least make it super obvious.


acsauk (approved):
Tidy :shipit: 

Can you make a 3 for 3 PR with a three-digit change count? Only time will tell...

PR #329, opened 2023-02-14
MLPAB-661: Certificate provider provides their details

acsauk:
# Purpose

Adds a couple of views to the CP flow to collect email, mobile, DOB and postal address.

Fixes MLPAB-661

## Approach

The bulk of this is extracting out some shared mocks and vars that are used in `donor` and `certificateprovider`. I've updated some of the var names so they're easier to grep in the future as it was a bit of a pain identifying which `appData` I wanted to find+replace.

I've dropped the lpa ID from the path for these pages as I *think* as part of an upcoming ticket we'll be relying on a one time code and then will store the ID in CP session. If that's not the plan I'll reintroduce the `/lpa` path back to these.

## Checklist

* [ ] I have performed a self-review of my own code


app/internal/page/certificateprovider/your_address.go
```diff
@@ -0,0 +1,85 @@
+package certificateprovider
+
+import (
+	"errors"
+	"net/http"
+
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/page"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/page/form"
```
hawx, line 8: What do you think about putting this at `github.com/ministryofjustice/opg-modernising-lpa/internal/form` instead? I was kind of thinking of keeping `page/*` as the different journeys. ...and later down the line we might move `validation` into or under `form`.

app/internal/page/certificateprovider/your_details.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/page"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/sesh"
+
+	"github.com/gorilla/sessions"
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/date"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/validation"
+)
+
+type cpYourDetailsData struct {
```
hawx, line 16: since these are now in a separate package we can call them `yourDetails...` and not have conflicts with the donor stuff

app/internal/page/donor/restrictions.go
```diff
@@ -29,15 +29,15 @@ func Restrictions(tmpl template.Template, lpaStore page.LpaStore) page.Handler {
 		}
 
 		if r.Method == http.MethodPost {
-			form := readRestrictionsForm(r)
```
hawx, line 32: you could have left these as `form`, it only becomes an issue when you have a variable called `form` in scope _and_ also want to use something from the package

app/internal/page/mock.go
```diff
@@ -0,0 +1,186 @@
...
+
+	"github.com/gorilla/sessions"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/identity"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/localize"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/onelogin"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/place"
+	"github.com/stretchr/testify/mock"
+)
+
+var MockRandom = func(int) string { return "123" }
```
hawx, line 20: I'm not super comfortable with having these in a common package since I prefer tests only to use things in their own package... but it is all just duplicated so I guess I can learn to like it. Maybe one day we can look into auto generating the mocks with https://vektra.github.io/mockery/

app/internal/page/certificateprovider/your_details.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		certificateProviderSession, err := sesh.CertificateProvider(sessionStore, r)
+		if err != nil {
+			return err
+		}
+
+		if certificateProviderSession.LpaID != lpa.ID {
+			return appData.Redirect(w, r, lpa, page.Paths.CertificateProviderStart)
+		}
```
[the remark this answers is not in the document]
acsauk: Nice - I was on the fence about moving this into `makeHandle()` but if it's taken care of further up the chain I'll drop out.

app/internal/page/certificateprovider/your_address.go
acsauk: That makes sense - I'll rejig out to its own internal package

app/internal/page/certificateprovider/your_details.go
acsauk: Missed these - will update.

app/internal/page/donor/restrictions.go
acsauk: Cool - I think I was running into issues in tests and just did a batch update. I do prefer the more descriptive var name so I'll see how reverting to `form` goes.

app/internal/page/mock.go
acsauk: Mmmm yeah I get where you're coming from but given the narrow scope of the stuff we're mocking it felt like needless repetition. If we get to a point where we need to deviate on the interfaces just hit me up for the grunt work undoing this 😅 

Interesting mocking package there, I'll take a look for future reference.

app/internal/page/mock.go
```diff
@@ -0,0 +1,197 @@
+package page
```
hawx, line 1: sorry missed the filename before, this should be `mock_test.go` so it doesn't get compiled into normal code

app/web/template/certificate_provider_your_details.gohtml
```diff
@@ -1,13 +1,36 @@
...
+          <fieldset class="govuk-fieldset">
+            <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+              <h1 class="govuk-fieldset__heading">{{ tr .App "yourDetails" }}</h1>
+            </legend>
+
+            <h2 class="govuk-heading-m">{{ tr .App "donor" }}</h2>
+            <p class="govuk-body">{{ .Lpa.You.FullName }}</p>
+
+            <h2 class="govuk-heading-m">{{ tr .App "yourDetails" }}</h2>
+            <p class="govuk-body">{{ .Lpa.CertificateProviderUserData.FullName }}</p>
```
hawx, line 19: I still don't think we should ever use the value from identity as their name, only as a source for matching logic. But could use `.Lpa.CertificateProvider.FullName` and same for `.DateOfBirth`, `.Email` and `.Mobile`?

app/web/template/certificate_provider_your_details.gohtml
```diff
@@ -1,13 +1,36 @@
...
-      <p class="govuk-body">Certificate provider is {{ .Lpa.CertificateProvider.FullName }}</p>
+      <form novalidate method="post">
+        <div class="govuk-form-group">
+          <fieldset class="govuk-fieldset">
+            <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+              <h1 class="govuk-fieldset__heading">{{ tr .App "yourDetails" }}</h1>
+            </legend>
+
+            <h2 class="govuk-heading-m">{{ tr .App "donor" }}</h2>
+            <p class="govuk-body">{{ .Lpa.You.FullName }}</p>
```
hawx, line 16: I only put this in to show it worked, probably don't want donor info on the "your details" page, but also don't want to get too sidetracked on the design of a page that will surely change...

app/web/template/certificate_provider_your_details.gohtml
acsauk: I'm thinking of instances where the Donor has got some name details wrong about the CP and then they've identified themselves but will still be presented with the incorrect name? I guess this comes into play more when we decide how CPs will request changes to LPA data and also what content/design want to end up showing on these templates.

I've purposefully collected the other details here following the comments on the skeleton - https://miro.com/app/board/uXjVPL-znGI=/?openComment=3458764545649811033

app/web/template/certificate_provider_your_details.gohtml
acsauk: I'll leave as is for now and we'll fix up when we get the real design for the page.

app/internal/page/mock.go
acsauk: I'd initially tried this but as soon as I postfix a `_test` on it the exported types and vars become unavailable in other packages. 

If we don't want this to be compiled is there some way of changing the build command to exclude directories/files?

app/internal/page/auth_redirect_test.go
```diff
@@ -1,68 +1,31 @@
...
-	args := m.Called(ctx, userInfo)
-	return args.Get(0).(identity.UserData), args.Error(1)
-}
-
 func TestAuthRedirect(t *testing.T) {
 	w := httptest.NewRecorder()
 	r, _ := http.NewRequest(http.MethodGet, "/?code=auth-code&state=my-state", nil)
 
-	client := &mockOneLoginClient{}
+	client := &MockOneLoginClient{}
```
hawx, line 20: these could be lowercase again

app/web/template/certificate_provider_your_details.gohtml
hawx: Since the name we get from ID may not be the same as you'd want to present yourself I don't think it should be used. I think it was a mistake for me to put `.FullName` on `UserData`, instead it should only give you the raw data plus some operations for (fuzzily) matching it.

hawx (approved):

app/web/template/certificate_provider_your_details.gohtml
hawx: I finally looked for a source https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1124033/Names_-_names_that_cannot_be_used_in_passports__V9_FOR_GOV.UK_publication_.pdf and page 24 of https://www.icao.int/publications/Documents/9303_p3_cons_en.pdf

hawx: also fun 

> The personal details section of the application allows the customer to use up to 30
characters for their forenames and 30 characters for their surname, including spaces
between them.

we allow more than that

hawx: and more https://deedpolloffice.com/change-name/restrictions-on-names, showing you can have a name which isn't equivalent to your "passport name", or even if you don't care about a passport you can break their rules

hawx: and something else to maybe consider for the Your Details page

>  There is no law preventing you from being known by a single name, or mononym — that is, a first name only, with no surname — and HM Passport Office should accept such a name, although they may be more sceptical of your application.

acsauk: Interesting - you've won me over. I'll drop a follow-up ticket in to rework the `UserData` stuff.

It would be really useful if we could get some example responses from GDS on names that contain diacritical characters or accent marks so we definitely know what we're working with here.

hawx (approved):

PR #330, opened 2023-02-15
MLPAB-669 Use a single code in link to certificate provider

hawx:
I took the key design from https://amazon-dynamodb-labs.workshop.aws/game-player-data/core-usage/step1.html, which seemed sensible enough. I was going to expire the code and add a different record to tie the certificate provider to the LPA but realised if you had multiple LPAs we'd then need a dashboard type thing to pick from, so I've gone the simple route of requiring the code each time on sign in. This can obviously be changed later.

### DIFF
--- a/app/internal/app/app.go
+++ b/app/internal/app/app.go
@@ -50,6 +50,7 @@ func App(
 		sessionStore,
 		lpaStore,
 		oneLoginClient,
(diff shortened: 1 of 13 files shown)


acsauk (approved):
:shipit:

PR #334, opened 2023-02-16
MLPAB-661: Certificate provider reads the LPA

acsauk:
# Purpose

Just defaulting to the existing template we have until we get any changes from content/design.

Fixes MLPAB-661.


acsauk:
Wrong branch 🤦

PR #340, opened 2023-02-17
MLPAB-523 Try to improve docker build

hawx:
The main thing is to not install `cypress` on build, looks to be around 20s faster for the build now. Moving things out of `app/` really only benefits local dev, but it is quite an improvement not to do a Go build when all that changed is a template.


Dockerfile
```diff
@@ -2,26 +2,20 @@ FROM node:18.14.0-alpine3.16 as asset-env
...
-COPY app/go.sum .
-
+COPY app/go.mod app/go.sum ./
 RUN go mod download
 
 COPY /app .
-
 RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -installsuffix cgo -o /go/bin/mlpab
 
 FROM build-env as development
```
acsauk, line 21: I don't think anyone is actually using air now so we can drop the whole `development` build env

hawx: :+1: cool

andrewpearce-digital (approved):

PR #346, opened 2023-02-20
MLPAB-377: Add fixtures page

acsauk:
# Purpose

Adds a fixtures page to allow setting donor flow state by section and links to jump into the CP flow either before or after the donor has IDed. This is aimed at helping non-devs/people-who-don't-like-query-strings to jump around the app when manual testing or quickly checking content.

Fixes MLPAB-377


app/internal/page/fixtures.go
```diff
@@ -67,3 +105,201 @@ func MakeCertificateProvider(firstNames string) actor.CertificateProvider {
...
+	lpa.Tasks.YourDetails = TaskCompleted
+}
+
+func AddAttorneys(lpa *Lpa, count int) []string {
+	if count > len(AttorneyNames) {
+		count = len(AttorneyNames)
+	}
+
+	var firstNames []string
+	for i := 0; i < count; i++ {
```
hawx, line 122: if you change those maps to slices these could be neat `for name := range AttorneyNames[:count] { ... }`

app/internal/page/fixtures.go
```diff
@@ -67,3 +105,201 @@ func MakeCertificateProvider(firstNames string) actor.CertificateProvider {
 		RelationshipLength:      "gte-2-years",
 	}
 }
+
```
hawx, line 108: I can't comment on the lines, but I've been getting paranoid about emails and mobile numbers, should we set them all to https://docs.notifications.service.gov.uk/rest-api.html#smoke-testing by default so people really can't mess up?

hawx (commented):
Having a UI for this will make it so much easier to explain

app/internal/page/fixtures.go
acsauk: Probably the safest bet yeah - I'll add a couple of consts in somewhere that we can reference where we need it in tests.

app/internal/page/fixtures.go
```diff
@@ -1,11 +1,49 @@
...
+	"golang.org/x/exp/slices"
+
+	"github.com/ministryofjustice/opg-go-common/template"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/actor"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/date"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/place"
 )
 
+var AttorneyNames = map[int]string{
```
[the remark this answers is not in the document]
acsauk: I was paranoid about the random order of slices when accessing them as the I'm relying on 0 and 1 being the names defined there to assert on - or am I getting this mixed up with another data structure that randomises things when accessed?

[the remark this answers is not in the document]
hawx: the only thing that randomises is the order of ranging over a map, accessing slices is fine they are just arrays (kinda)

hawx (approved):

hawx (approved):

PR #349, opened 2023-02-21
MLPAB-679 Generate mocks

hawx:
This is only the simple mocks. Having them written for us with auto-asserting is pretty nice. I'll tackle the weirder ones in a different PR, and worst case manually add the same kind of auto-asserting to them.

I've added `make go-generate` to run the generation, which we'll need to do whenever we add a new mock or method to a mocked interface. I don't think it's worth being paranoid about checking in the GitHub Actions that this has been done because the tests will fail if the mock doesn't match the interface.


acsauk (approved):
:shipit:

PR #350, opened 2023-02-21
MLPAB-679 Generate the weird mocks

hawx:
Adding our own methods on them seems to work quite nicely


app/internal/page/certificateprovider/mock_Template_test.go
```diff
@@ -0,0 +1,43 @@
...
+	mock "github.com/stretchr/testify/mock"
+)
+
+// mockTemplate is an autogenerated mock type for the Template type
+type mockTemplate struct {
+	mock.Mock
+}
+
+// Execute provides a mock function with given fields: _a0, _a1
+func (_m *mockTemplate) Execute(_a0 io.Writer, _a1 interface{}) error {
```
acsauk, line 17: Is this what is eventually being called when `template.Func` is run in our handlers?

hawx: yep, it works just like `Func` did before

acsauk (approved):
:shipit:

PR #352, opened 2023-02-23
MLPAB-694: Enter reference code

acsauk:
# Purpose

Adds a page to enter the share code generated when inviting the CP.

Fixes MLPAB-694.

## Approach

I've updated the  CP flow fixture buttons to actually send the email with a static sharecode in so we have a way to step through the flow when testing with users.


app/internal/page/certificateprovider/enter_reference_code.go
```diff
@@ -0,0 +1,82 @@
...
+			data.Errors = data.Form.Validate()
+
+			if len(data.Errors) == 0 {
+				shareCode := data.Form.ReferenceCode
+
+				var v page.ShareCodeData
+				if err := dataStore.Get(r.Context(), "SHARECODE#"+shareCode, "#METADATA#"+shareCode, &v); err != nil {
+					return err
+				}
+
```
hawx, line 38: If the sharecode can't be retrieved we should show a validation error because it is probably just a mistype. (I can't recall why I made dynamo.Client.Get return a `nil` error if the item couldn't be retrieved, but that is what it currently does)

app/internal/page/share_code.go
```diff
@@ -7,30 +7,43 @@ import (
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/notify"
 )
 
+var useTestCode = false
```
hawx, line 10: we need to come up with a different way of doing this (and the one in `random`) as it will be set `true` until the app is restarted, making using the app "properly" impossible if someone uses testing-start

acsauk: Is the issue that we inject `sharecode` at run time and then can't change it? I'd hoped by having the property here (and assuming we're passing round a pointer) we could update it.

Maybe we need `Send()` to always reset `useTestCode` back to false after it's been set to true, like:

```golang
var shareCode string

if useTestCode {
    shareCode = "abcdef123456"
    useTestCode = false
} else {
    shareCode = s.randomString(12)
}
```

app/internal/page/certificateprovider/enter_reference_code.go
acsauk: I did have incorrect codes at the back of my head but was just falling back on the error - hadn't realised `Get()` was defaulting to a nil error. Agree having a validation error would be nicer though - I'll take a look.

app/internal/page/certificateprovider/enter_reference_code_test.go
```diff
@@ -0,0 +1,209 @@
...
+		On("Execute", w, data).
+		Return(nil)
+
+	err := EnterReferenceCode(template.Execute, newMockLpaStore(t), newMockDataStore(t))(testAppData, w, r)
+
+	resp := w.Result()
+
+	assert.Nil(t, err)
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+	mock.AssertExpectationsForObjects(t, template)
```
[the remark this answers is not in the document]
acsauk: Nice - I wasn't sure if we had to do some setup with the mock but getting it for free is 👌

app/internal/page/share_code.go
hawx: yeah that should work for now

app/internal/dynamo/client.go
```diff
@@ -63,7 +69,7 @@ func (c *Client) Get(ctx context.Context, pk, sk string, v interface{}) error {
 		return err
 	}
 	if result.Item == nil {
-		return nil
+		return &NotFoundError{}
```
hawx, line 72: since you defined `Error` on the non-pointer receiver you could have returned `NotFoundError{}` here

hawx (approved):

PR #356, opened 2023-02-24
MLPAB-639 Correct the nesting for govuk-form-group and govuk-fieldset

hawx:
- removed outer form-groups that weren't doing anything
- removed fieldsets wrapping multiple unrelated controls
- put form-group--errors outside fieldsets so the red error line includes the legend


acsauk (approved):
:shipit:

PR #360, opened 2023-02-28
MLPAB-722 Split certificate provider sign in and identity

hawx:
I'm doing this the copy-paste way because I don't trust people know what they want the same and what they don't yet, so it will probably be better to decide what to do with it all later.


web/template/certificate_provider_your_details.gohtml
```diff
@@ -13,7 +13,7 @@
         <p class="govuk-body">{{ .Lpa.You.FullName }}</p>
 
         <h2 class="govuk-heading-m">{{ tr .App "yourDetails" }}</h2>
-        <p class="govuk-body">{{ .Lpa.CertificateProviderUserData.FullName }}</p>
+        <p class="govuk-body">{{ .Lpa.CertificateProvider.FullName }}</p>
```
hawx, line 16: since we won't now have ID info at this point

acsauk (approved):
:shipit:

PR #364, opened 2023-03-01
Use donor's email as default for payment

hawx:
After seeing this in the demo for the Nth time, I figured it should really be doing this


acsauk (approved):
:shipit:

PR #367, opened 2023-03-01
MLPAB-650: Hide People to Notify summary actions when LPA signed

acsauk:
# Purpose

Updates template logic to check for the LPA being signed rather than just checked.

Fixes MLPAB-650

## Approach

I took the opportunity to update the PTN e2e tests as they were sometimes taking 10 seconds to run in CI. This should speed things up a little. Also adds an input for specifying how many PTN to add in fixtures - if there's a need for this in other actor form sections we can replicate.


app/internal/page/fixtures_test.go
```diff
@@ -57,7 +58,7 @@ func TestPostFixturesDonorFlow(t *testing.T) {
 
 	assert.Nil(t, err)
 	assert.Equal(t, http.StatusFound, resp.StatusCode)
-	expectedPath := "/testing-start?completeLpa=1&idConfirmedAndSigned=1&lpaChecked=1&paymentComplete=1&whenCanBeUsedComplete=1&withAttorneys=1&withCP=1&withDonorDetails=1&withPeopleToNotify=1&withReplacementAttorneys=1&withRestrictions=1"
+	expectedPath := "/testing-start?completeLpa=1&idConfirmedAndSigned=1&lpaChecked=1&paymentComplete=1&peopleToNotifyCount=5&whenCanBeUsedComplete=1&withAttorneys=1&withCP=1&withDonorDetails=1&withPeopleToNotify=1&withReplacementAttorneys=1&withRestrictions=1"
```
hawx, line 61: could have just made the `withPeopleToNotify` param be the count

web/template/layout/people-to-notify-summary.gohtml
```diff
@@ -18,7 +18,7 @@
                 <dd class="govuk-summary-list__value">
                     {{ $p.FirstNames }} {{ $p.LastName }}
                 </dd>
-                {{ if not (eq $.Lpa.Tasks.CheckYourLpa.String "completed") }}
+                {{ if not (eq true $.Lpa.Tasks.ConfirmYourIdentityAndSign.Completed) }}
```
hawx, line 21: since it is a boolean `eq true X` is the same as `X`

web/template/fixtures.gohtml
```diff
@@ -55,11 +55,19 @@
...
-                People to notify about your LPA (adds 2)
+                People to notify about your LPA
               </label>
             </div>
+            <div class="govuk-checkboxes__conditional" id="conditional-ptn-count">
+              <div class="govuk-form-group">
+                <label class="govuk-label" for="f-ptn-count">
+                  How many? (Max 5)
+                </label>
+                <input class="govuk-input govuk-!-width-one-third" id="f-ptn-count" name="ptn-count">
```
hawx, line 68: maybe give this a value of 2 to show the default

app/internal/page/fixtures_test.go
acsauk: I had a really valid reason in my head for doing it this way yesterday that has now vanished. Makes sense to have one less query for sure, I'll switch it out.

hawx (approved):

PR #369, opened 2023-03-03
Rename You to Donor on Lpa

hawx:
The original intention was that we'd do something smart based on whether you picked "yourself" or "someone else" at the start, but we are just spreading `You` around as being the donor. This rename will make it clearer when we get to that feature that nothing smart has been done.


acsauk (approved):
:shipit:

PR #377, opened 2023-03-07
MLPAB-550 Rate limit entry of the witness code

hawx:
I felt that embedding the rate limiter state in the Lpa was going to be nicer than having a global map of limiters (that would then require cleaning up, and wouldn't scale if we had more than 1 container).

Also missed that I'd changed some content in my previous PR copy and pasting from the GDS pattern, but not reverted it.


acsauk (approved):
:shipit:

PR #380, opened 2023-03-08
Remove unused data

hawx:
Since I went through what we store I spotted some things that aren't used now


app/internal/page/certificateprovider/check_your_name_test.go
```diff
@@ -7,12 +7,10 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/ministryofjustice/opg-modernising-lpa/internal/notify"
-
-	"github.com/ministryofjustice/opg-modernising-lpa/internal/validation"
-
```
acsauk, line 13: Haha still not sent that message to JetBrains

acsauk (approved):
:shipit:

PR #386, opened 2023-03-09
MLPAB-765 Match names retrieved from the identity checks

hawx:
This consolidates the identity info into one field for the donor and one for the certificate provider. I've then expanded the `Lpa.IdentityConfirmed` method (and added a `Lpa.CertificateProviderIdentityConfirmed` method) to check the names of the actor match the data returned. This is first making everything uppercase, then if they don't match making the order of firstnames not matter and transliterating characters using the data from the table in the comments. I have not gone so far as to do alternate transliterations as that will require "actual work" and we aren't sure yet how far this needs to be taken.

I have then also made it so that if identity has been confirmed you will skip to "Read the LPA" from the task list, and blocked progress to some pages if identity has not been confirmed.

### DIFF
--- a/app/internal/page/data.go
+++ b/app/internal/page/data.go
@@ -66,9 +66,8 @@ type Lpa struct {
 	Checked                                     bool
 	HappyToShare                                bool
 	PaymentDetails                              PaymentDetails
-	IdentityOption                              identity.Option
-	YotiUserData                                identity.UserData
-	OneLoginUserData                            identity.UserData
+	DonorIdentityOption                         identity.Option
+	DonorIdentityUserData                       identity.UserData
 	HowAttorneysMakeDecisions                   string
 	HowAttorneysMakeDecisionsDetails            string
 	ReplacementAttorneys                        actor.Attorneys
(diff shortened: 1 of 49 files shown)


app/internal/page/data.go
```diff
@@ -141,7 +139,13 @@ func DecodeAddress(s string) *place.Address {
 }
 
 func (l *Lpa) IdentityConfirmed() bool {
```
acsauk, line 141: Seeing as we've now switched from `Person` to `Donor` is it worth scoping this func name to `DonorIdentityConfirmed` for clarity?

app/internal/page/donor/identity_with_yoti_callback_test.go
```diff
@@ -40,19 +52,84 @@ func TestGetIdentityWithYotiCallback(t *testing.T) {
...
+				CouldNotConfirm: true,
+			}).
+			Return(nil)
+		return template
+	}
+
+	templateIgnored := func(t *testing.T, w io.Writer) *mockTemplate {
+		return nil
+	}
+
```
acsauk, line 70: Neat 👍

acsauk (commented):
One comment on a func name but good to go other than that - looking forward to getting some users with many accented letters in their name in UR 😁

andrewpearce-digital (approved):

PR #387, opened 2023-03-09
MLPAB-750 - fix path to live

andrewpearce-digital:
# Purpose

Fix path to live

Get quick feedback on jobs without having to delve into the actions log

Fixes MLPAB-750

## Approach

- pipe a summary of the plan to an output
- use actions scripts to push a message to PR (or any other issue type) comments

## Learning

- https://github.com/actions/github-script#comment-on-an-issue
- https://github.com/hashicorp/setup-terraform/blob/main/README.md

To capture the plan summary:
The `-no-color` option is used to disable colourised output from the terraform plan command, and the `-ioE` options for grep are used to perform a case-insensitive search (-i), to output only the matching text (-o), and to use extended regular expressions (-E) for matching the numbers in the summary line.

## Checklist

### DIFF
--- a/.github/workflows/workflow_path_to_live.yml
+++ b/.github/workflows/workflow_path_to_live.yml
@@ -9,11 +9,11 @@ permissions:
   id-token: write
   contents: write
   security-events: write
-  pull-requests: read
+  pull-requests: write
   actions: none
   checks: none
   deployments: none
-  issues: none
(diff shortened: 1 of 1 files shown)


johnnolan (approved):

acsauk (approved):
:shipit:

PR #395, opened 2023-03-13
MLPAB-785 Fix yoti redirect

hawx:
I think that having a single scenario like this will make life easier, since we want the same data from both actors. I'm using the same pattern as we do for One Login, storing data in a session and adding a facade redirect handler.


acsauk (commented):
One potential change for a var but otherwise good to go :shipit:

PR #398, opened 2023-03-16
MLPAB-593: Update witness code personalisations and notify template IDs

acsauk:
# Purpose

And adds a possessive-ication func on Donor so we can treat last names ending with S correctly.

Fixes MLPAB-593


app/internal/actor/donor.go
```diff
@@ -16,6 +16,26 @@ type Donor struct {
...
+
+		if d.FirstNames[len(d.FirstNames)-1:] == "s" {
+			format = "%s’"
+		}
+
+		return fmt.Sprintf(format, d.FirstNames)
+	} else {
+		format := "%s %s’s"
+
+		if d.LastName[len(d.LastName)-1:] == "s" {
```
hawx, line 35: maybe `strings.HasSuffix(d.LastName, "s")` is clearer

hawx: and you could cut down the duplicated logic with something like this at the start

```go
name := d.LastName
if d.LastName == "" {
  name = d.FirstNames
}
```

hawx: Also, I wonder how this all works in Welsh... https://old.reddit.com/r/learnwelsh/comments/5oqafl/welsh_grammar_the_genitivepossessive_construction/

Perhaps for now, so we remember it doesn't actually work, pass the language in and for Cy just return "Welsh" like we do for translations?

app/internal/page/witness_code.go
```diff
@@ -30,15 +30,17 @@ func NewWitnessCodeSender(lpaStore LpaStore, notifyClient NotifyClient) *Witness
 	}
 }
 
-func (s *WitnessCodeSender) Send(ctx context.Context, lpa *Lpa) error {
+func (s *WitnessCodeSender) Send(ctx context.Context, lpa *Lpa, appData AppData) error {
```
hawx, line 33: could just pass the Localizer through, since that's all you need

hawx (approved):

PR #411, opened 2023-03-21
MLPAB-799 - Toggle public access for UR environment

andrewpearce-digital:
# Purpose

Make it easy for team members to toggle UR availability using github actions

Fixes MLPAB-799

## Approach

- split public access variable into separate var
- create a job to toggle public access from github actions
- manage public access for production env specifically

## Notes

- Configuring this to target a single environment that isn't ur for now while I test

## Checklist

* [] I have performed a self-review of my own code

### DIFF
--- a/.github/workflows/workflow_manage_public_access_ur_environment.yml
+++ b/.github/workflows/workflow_manage_public_access_ur_environment.yml
@@ -37,4 +37,3 @@ jobs:
         aws_access_key_id: ${{ secrets.AWS_ACCESS_KEY_ID_ACTIONS }}
         aws_secret_access_key: ${{ secrets.AWS_SECRET_ACCESS_KEY_ACTIONS }}
         ssh_deploy_key: ${{ secrets.OPG_MODERNISING_LPA_DEPLOY_KEY_PRIVATE_KEY }}
-        github_access_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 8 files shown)


.github/workflows/workflow_manage_public_access_ur_environment.yml
```diff
@@ -0,0 +1,39 @@
...
+  run:
+    shell: bash
+
+jobs:
+  ur_toggle_public_access:
+      name: UR Environment Deploy
+      needs: [ui_tests_image]
+      uses: ./.github/workflows/toggle_public_access_job.yml
+      with:
+        workspace_name: 411mlpab799
```
hawx, line 34: should it be this?

andrewpearce-digital: yes, while I'm testing I don't want to modify the ur env (it's also in use at the moment). While I've tested the commands etc. I also can't access this workflow dispatch while it's in the PR. Once it's merged I'll be able to access the job (and branch versions of it too)

hawx (approved):

.github/workflows/workflow_manage_public_access_ur_environment.yml
andrewpearce-digital: yes, I don't want to modify UR while it's in use. While this is a PR, the workflow dispatch job doesn't appear for me to trigger and test (I've tested the commands etc).

PR #418, opened 2023-03-23
MLPAB-776: Re-use existing donor, attorney, replacement attorney or CP addresses

acsauk:
# Purpose

I've kept access to this page from the initial enter address screen. I'm sure there are other places we may want to use this but as this hasn't been through design and content yet it's worth waiting for their input.

Fixes MLPAB-776

## Approach

I was on the verge of implementing some high-level interface for an actor but as there are some unknowns on how this feature was going to work I didn't want to start swapping out argument signatures. It would have meant we could avoid the `AddressDetail` type (which is hyper-focused on its use on this page) and just worked with a list of actors. Something to think about for the future maybe.


app/internal/page/data.go
```diff
@@ -225,3 +226,46 @@ type Certificate struct {
...
+type AddressDetail struct {
+	Name    string
+	Role    actor.Type
+	Address place.Address
+	ID      string
+}
+
+func (l *Lpa) ActorAddresses() []AddressDetail {
+	var ads []AddressDetail
+
```
hawx, line 239: and the donor's address?

hawx (dismissed):

app/internal/page/data.go
hawx: although maybe don't care too much, since this is more about putting something together to look at

acsauk: I'd originally left this out after seeing a screen of the same address between actors but this was before I'd added the actor type in the template. I can definitely see the value now so will add back in 👍

hawx (approved):

PR #423, opened 2023-03-30
MLPAB-790 Correct replacement attorney flows

hawx:


app/internal/page/donor/remove_attorney.go
```diff
@@ -44,6 +44,10 @@ func RemoveAttorney(logger Logger, tmpl template.Template, lpaStore LpaStore) pa
 
 			if data.Form.RemoveAttorney == "yes" && data.Errors.None() {
 				lpa.Attorneys.Delete(attorney)
+				if len(lpa.Attorneys) == 1 {
```
hawx, line 47: Doing this means I don't have to check that there are multiple attorneys in other places. But I think I'm going to play around with refactoring `Attorneys` to include the decisions so all the logic will be in one place, but in a different PR.

app/internal/page/data.go
```diff
@@ -314,6 +314,12 @@ func ChooseReplacementAttorneysState(lpa *Lpa) TaskState {
 		}
 	}
 
+	if len(lpa.ReplacementAttorneys) > 1 &&
+		lpa.HowShouldReplacementAttorneysStepIn != OneCanNoLongerAct &&
```
hawx, line 318: just waiting on a decision for if this should be `== AllCanNoLongerAct` instead

cypress/e2e/donor/choose-replacement-attorneys-task.cy.js
```diff
@@ -13,7 +13,7 @@ describe('Choose replacement attorneys task', () => {
 
         cy.contains('label', 'No').click();
         cy.contains('button', 'Continue').click();
-        
+
```
hawx, line 16: I'm thinking about keeping these but skipping them? They are really useful tests for documenting this flow when trying to debug stuff, as there are just too many pages to keep track of, but they are slow.

cypress/e2e/donor/choose-attorneys-task.cy.js
```diff
@@ -1,6 +1,6 @@
 import {TestEmail} from "../../support/e2e";
 
-describe('Choose attorneys task', () => {
+describe.skip('Choose attorneys task', () => {
```
hawx, line 3: I think it's nice to have some tests that document these flows, because there are too many. But they are slow. So skipping them seems like a sensible middle ground?

acsauk: Yeah good compromise 👍

acsauk (approved):
:shipit:

PR #456, opened 2023-04-25
MLPAB-882 - Updates to maintenance mode job

andrewpearce-digital:
# Purpose

Make actions jobs easier to use

Fixes MLPAB-882

## Approach

- add dispatch, job and workflow names to make working with dispatch jobs easier to identify
- fix how outputs are set


acsauk (approved):
:shipit:

PR #461, opened 2023-04-27
MLPAB-730: Add read the LPA page to attorney flow

acsauk:
# Purpose

Adds read the LPA page for attorneys and some extra checks in cypress tests for varying actor-related content.

Fixes MLPAB-730


hawx (commented):
can you add this into the page flow, and link from the task list?

hawx (approved):

PR #463, opened 2023-04-28
MLPAB-857 - Add job to switch between deployed regions

andrewpearce-digital:
# Purpose

Allow team to switch between eu-west-1 and eu-west-2 regions when needed

Fixes MLPAB-857

## Approach

Before the DNS target can be changed, the second region account resources and environment resources must be deployed.

- add a parameter store value to configure DNS target region
- add job to update parameter store value for DNS target region
- ~run a targeted apply to update the DNS target region~ This is commented out for this PR until we have second region resources

As usual, I need to merge in the dispatch job before I can test using it. Fixes and improvements will come in a later PR.

## Checklist

### DIFF
--- a/.github/workflows/dispatch_change_dns_target_region.yml
+++ b/.github/workflows/dispatch_change_dns_target_region.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       environment:
-        description: 'Environment to to update DNS target region for'
+        description: 'Environment to update DNS target region for'
         required: true
         type: string
       region:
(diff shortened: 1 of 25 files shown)


.github/workflows/dispatch_change_dns_target_region.yml
```diff
@@ -0,0 +1,92 @@
+name: "[WD] Change DNS Target Region"
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: 'Environment to to update DNS target region for'
```
acsauk, line 7:
```suggestion
        description: 'Environment to update DNS target region for'
```

(as there's a follow up coming may be worth tackling in that)

acsauk (approved):
One typo but that can be fixed on the follow up ticket 👍

PR #464, opened 2023-04-28
MLPAB-904 Handle witness code submission when certificate provider has no data

hawx:


acsauk (approved):
My bad! Thanks for sorting 👍 

:shipit:

PR #465, opened 2023-04-28
MLPAB-899: Add setup for local app debug

acsauk:
# Purpose

Got there in the end 💪 I've added some toggling for processor architecture as I hit some issues building air and/or delve under amd64. Hopefully, this doesn't break anyone's local setups but shout up if so.

Fixes MLPAB-899.


web/template/attorney_what_happens_next.gohtml
```diff
@@ -16,6 +16,7 @@
           {{ end }}
         </h1>
       </div>
+      asasas
```
hawx, line 19: :wave:

hawx (approved):

PR #472, opened 2023-05-04
MLPAB-732: Add what happens when you sign page

acsauk:
# Purpose

I got tired of printing `Lpa{}` props in templates to check if I'd set cypress tests up right so I've added a tmplfn for easier dumping of structs in templates. 

Fixes MLPAB-732


app/internal/page/data.go
```diff
@@ -364,3 +364,8 @@ func ChooseReplacementAttorneysState(lpa *Lpa) TaskState {
 
 	return TaskCompleted
 }
+
+// TODO update references in code from hw to pw and pfa to pa
```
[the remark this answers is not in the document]
hawx: also it is much easier to see the difference between hw/pfa than pa/pf

hawx (dismissed):

app/internal/page/data.go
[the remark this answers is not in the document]
acsauk: Fair point - I'll switch this to `IsHealthAndWelfareLpa()`

web/template/attorney_what_happens_when_you_sign.gohtml
```diff
@@ -7,9 +7,11 @@
         <div class="govuk-grid-column-two-thirds">
             <h1 class="govuk-heading-xl">{{ tr .App "whatHappensWhenYouSignTheLpaTitle" }}</h1>
 
+            {{ printf "%#v\n" .Lpa }}
```
hawx, line 10: you'll want to remove this

acsauk: Ugh - thanks, dropped now 👍

app/internal/page/attorney/register.go
```diff
@@ -111,8 +111,10 @@ func Register(
 		YourAddress(logger, tmpls.Get("your_address.gohtml"), addressClient, lpaStore))
 	handleRoot(page.Paths.Attorney.ReadTheLpa, RequireLpa,
 		ReadTheLpa(tmpls.Get("attorney_read_the_lpa.gohtml"), lpaStore))
-	handleRoot(page.Paths.Attorney.RightsAndResponsibilities, RequireSession,
+	handleRoot(page.Paths.Attorney.RightsAndResponsibilities, None,
```
hawx, line 114: I think this should `RequireSession`, or the continue button won't work. Also we already have a general rights and responsibilities page linked from the footer that sits outside the flows.

PR #473, opened 2023-05-05
MLPAB-888 Update content of errors for attorneys

hawx:


app/internal/validation/check.go
```diff
@@ -75,27 +75,46 @@ func (l *List) Options(name, label string, value []string, checks ...OptionsChec
...
 
-type SelectedCheck struct{}
+type SelectedCheck struct {
+	useCustomError bool
+}
+
+func (c SelectedCheck) CustomError() SelectedCheck {
+	c.useCustomError = true
+	return c
+}
```
acsauk, line 85: Nice compromise 👍

acsauk (approved):
:shipit:

app/internal/validation/check.go
hawx: Yeah, hopefully it doesn't end up spreading too far though

PR #475, opened 2023-05-09
MLPAB-912 - Fixes to second region deployment

andrewpearce-digital:
# Purpose

Deploy the second region resources

Fixes MLPAB-912

## Approach

- Do not create an S3 ACL as they are no longer enabled/allowed for bucket by default (the default permissions are the same as the "private" ACL. This is suitable for the use case.
- ensure that replicated KMS keys also have the correct policies attached

## Learning

https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html

## Checklist


acsauk (approved):
:shipit:

PR #479, opened 2023-05-10
MLPAB-874: Move eligibility page before log in

acsauk:
# Purpose

Moves the CP eligibility page prior to login after enter your reference number.

As the CP will have to navigate through two screens before being redirected to one login I decided to wrap the LPA ID we get from the share code in a cookie in case the user decides to go back and forth and we end up losing the query params.

Fixes MLPAB-874


app/internal/page/certificateprovider/who_is_eligible.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/validation"
+)
+
+type whoIsEligibleData struct {
+	App    page.AppData
+	Lpa    *page.Lpa
+	Errors validation.List
+}
+
+func WhoIsEligible(tmpl template.Template, lpaStore LpaStore, store sesh.Store) page.Handler {
```
acsauk, line 18: Annoyingly needed this handler as `Guidance()` requires a valid session to work.

hawx (approved):

PR #511, opened 2023-06-02
MLPAB-944 Navigation updates

hawx:
The backlink uses JS to mimic the back button, so noscript people won't get one.

### DIFF
--- a/web/assets/main.js
+++ b/web/assets/main.js
@@ -1,7 +1,7 @@
-import * as GOVUKFrontend from "govuk-frontend";
-import $ from 'jquery'
-import { initAll } from '@ministryofjustice/frontend'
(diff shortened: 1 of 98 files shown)


andrewpearce-digital (approved):

web/assets/main.js
```diff
@@ -14,31 +14,39 @@ if (header) {
...
 }
 
+const backLink = document.querySelector('.govuk-back-link');
+if (backLink) {
+    backLink.addEventListener('click', function (e) {
+        window.history.back();
+        e.preventDefault();
+    }, false);
+}
+
```
acsauk, line 24: Cool, I hadn't seen this functionality in the past. SAves us from having to roll our own URL saving.

PR #516, opened 2023-06-06
MLPAB-968: Cover UID client in pact tests and send to pact broker

acsauk:
# Purpose

Adds a success and non-success contract test for the integration with the UID service.

Fixes MLPAB-968


andrewpearce-digital (approved):
I'm happy with all the actions related stuff

hawx (approved):

PR #517, opened 2023-06-07
MLPAB-894 Nest non-donor routes

hawx:
As we're going to be allowing people to freely move between viewing an LPA as a donor, attorney or certificate provider we need to put the LpaID in the URL as opposed to the session (which won't change after signing in).

### DIFF
--- a/app/internal/app/app.go
+++ b/app/internal/app/app.go
@@ -93,6 +93,7 @@ func App(
 		yotiClient,
 		notifyClient,
 		certificateProviderStore,
+		notFoundHandler,
 	)
 
 	attorney.Register(
@@ -108,6 +109,7 @@ func App(
 		shareCodeStore,
 		errorHandler,
 		notifyClient,
+		notFoundHandler,
 	)
 
 	donor.Register(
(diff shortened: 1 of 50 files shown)


app/internal/actor/attorney_provided.go
```diff
@@ -8,6 +8,7 @@ import (
 )
 
 type AttorneyProvidedDetails struct {
+	ID            string
```
hawx, line 11: now need this so we can go from `sub` to who they are on the Lpa

app/internal/page/attorney/guidance.go
```diff
@@ -14,8 +15,8 @@ type guidanceData struct {
 	Lpa    *page.Lpa
 }
 
-func Guidance(tmpl template.Template, donorStore DonorStore) page.Handler {
-	return func(appData page.AppData, w http.ResponseWriter, r *http.Request) error {
+func Guidance(tmpl template.Template, donorStore DonorStore) Handler {
+	return func(appData page.AppData, w http.ResponseWriter, r *http.Request, _ *actor.AttorneyProvidedDetails) error {
```
acsauk, line 19: I've not seen `_` used as an argument name before. Does this work in the same way as var names to show its unused?

hawx: yep, just showing it isn't used

acsauk (approved):
Nice improvement not having to initially get the attorney in the Handler code 🙌 can that pattern work for CPs and Donors?

PR #522, opened 2023-06-09
MLPAB-986: Allow donor to progress past /lpa-type on UID client errors

acsauk:
# Purpose

And retry the POST on `/choose-attorneys` (future PR will update all other LPA paths with the new handler so we retry the UID service on each request).

Fixes MLPAB-986.


app/internal/page/donor/register.go
```diff
@@ -329,3 +334,70 @@ func makeHandle(mux *http.ServeMux, store sesh.Store, defaultOptions handleOpt,
...
+				ctx = page.ContextWithSessionData(ctx, &page.SessionData{SessionID: appData.SessionID, LpaID: appData.LpaID})
+			}
+
+			lpa, err := donorStore.Get(ctx)
+			if err != nil {
+				errorHandler(w, r, err)
+				return
+			}
+
+			if lpa.Tasks.YourDetails == actor.TaskCompleted && lpa.UID == "" {
```
hawx, line 375: it might also be good to have `r.Method == http.MethodPost` here, as it might be weird seeing an (unlikely) error from the donorStore write

hawx (dismissed):

hawx (approved):

PR #523, opened 2023-06-12
MLPAB-1057 - set env vars for both destory jobs where they need a pagerduty api key

andrewpearce-digital:
# Purpose

Scheduled destroy job is failing due to missing input variable

Fixes MLPAB-1057

## Approach

- Add input variable to environment


acsauk (approved):
:shipit:

PR #524, opened 2023-06-12
MLPAB-963 - refactor regional policies for execution role to region module

andrewpearce-digital:
# Purpose

Clean up regional resources in root of environment configuration

Fixes MLPAB-963

## Approach

- Move regional secrets for RUM to region module
- Create regional policy documents and attach them to the ECS execution role


acsauk (approved):
:shipit:

PR #527, opened 2023-06-13
MLPAB-1061: Fix UID service expired token error

acsauk:
# Purpose

When signing requests using aws v4, the token that is added to the header has an expiry date. The code to generate the header was correct but as we retrieved the credentials from AWS on app startup in main it meant they were only valid for a short period of time.

This change moves requesting credentials alongside signing the request to ensure the token will always be valid from a time perspective.

Fixes MLPAB-1061


andrewpearce-digital (approved):

app/internal/page/dependency_health_check.go
```diff
@@ -0,0 +1,26 @@
...
+				Postcode: "B147ED",
+			},
+		})
+
+		if err != nil {
+			logger.Print(err)
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+	}
+}
```
acsauk, line 26: I've not added any tests to call this just yet as I'm thinking ahead to when prod is plumbed in we don't want to be creating dummy data in Sirius. I'm going to suggest a ping endpoint in the UID service that will at least give us confidence our signing and env config is correct without having to create data. Once that's working we can swap this out for a call to the new endpoint.

andrewpearce-digital (approved):

PR #528, opened 2023-06-13
MLPAB-1064 - Refactor DNS

andrewpearce-digital:
# Purpose

Make each region responsible for it's DNS and use weighted routing to distribute across or migrate to a region

Fixes MLPAB-1064

## Approach

- Migrate DNS record to region and add weighted routing policy
- Migrate DNS health check


acsauk (approved):
:shipit:

PR #529, opened 2023-06-13
MLPAB-1039 - Replace Config File with Outputs

andrewpearce-digital:
# Purpose

Remove need for a file that is moved around in favour of job outputs

Fixes MLPAB-1039

## Approach

- output environment config as json
- pass output to ingress management jobs
- remove config file
- ~mask json so that it doesn't appear in the actions logs~

At this time unable to mask outputs across jobs/workflows. A later ticket will explore putting the environment config into secrets manager

## Learning

- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-masking-and-passing-a-secret-between-jobs-or-workflows


acsauk (dismissed):
:shipit:

hawx (approved):

PR #531, opened 2023-06-15
MLPAB-1039 - fix path to live and ur workflows

andrewpearce-digital:
# Purpose

Fix failure on path to live and the same error that would occur on the ur workflow

Fixes MLPAB-1039

## Approach

- correct the name of the job that the output is taken from when managing environment ingress

## Learning

## Checklist

* [x] Changes to Github Actions jobs have been checked for all workflows


acsauk (approved):
:shipit:

PR #536, opened 2023-06-16
MLPAB-894 Reorganise fixtures and testing start

hawx:


.codecov.yml
```diff
@@ -7,6 +7,8 @@ coverage:
     - "./mocks/*"
     - "./app/internal/identity/yoti*"
     - "./app/internal/telemetry"
+    - "./app/internal/page/fixtures.go"
+    - "./app/internal/page/testing_start.go"
```
hawx, line 11: I don't see the value in testing these as `fixtures` doesn't really do much (and people actually use this page) and if `testing_start` broke we'd see it in the cypress tests

app/internal/page/testing_start.go
```diff
@@ -16,159 +18,390 @@ import (
...
-	return func(w http.ResponseWriter, r *http.Request) {
-		var (
-			donorSub                     = randomString(16)
-			donorSessionID               = base64.StdEncoding.EncodeToString([]byte(donorSub))
-			certificateProviderSub       = randomString(16)
-			certificateProviderSessionID = base64.StdEncoding.EncodeToString([]byte(certificateProviderSub))
-			attorneySub                  = randomString(16)
-			attorneySessionID            = base64.StdEncoding.EncodeToString([]byte(attorneySub))
-		)
+	const (
```
hawx, line 21: I moved the setup stuff into here to hide it from the "real" code

.codecov.yml
acsauk: Cool, makes sense

acsauk (approved):
:shipit:

PR #541, opened 2023-06-20
MLPAB-894 Formattable paths

hawx:
The idea is that `.String` returns just the end path component, so suitable for using in routers, and in basically every other situation we should use `.Format` so we can't miss the params.


acsauk (approved):
Nice and neat now 💪 

:shipit:

PR #545, opened 2023-06-22
MLPAB-865 Update certificate provider details content

hawx:


app/internal/actor/certificate_provider.go
```diff
@@ -16,6 +16,8 @@ type CertificateProvider struct {
 	Address place.Address
 	// Mobile number of the certificate provider, used to send witness codes
 	Mobile string
+	// HasNonUKMobile indicates whether the value of Mobile is a non-UK mobile number
+	HasNonUKMobile bool
```
hawx, line 20: we aren't going to be using the number any differently so decided to store it in `Mobile`, we just need this so the UI shows the checkbox correctly when going back to the page

acsauk (approved):
:shipit:

PR #553, opened 2023-06-26
MLPAB-1095 - upgrade aws provider to v5.5.0

andrewpearce-digital:
# Purpose

Upgrade provider

Fixes MLPAB-1095

## Approach

- update requirements for all modules
- update network module which has been updated to use v5.5.0

Should result in zero diffs


acsauk (approved):
:shipit:

PR #555, opened 2023-06-27
MLPAB-981: Add data loss warning dialog

acsauk:
# Purpose

Heavily borrowed from [Use an LPA](https://github.com/ministryofjustice/opg-use-an-lpa/blob/b4264a984477ecdbabedc7e1be873daf4fc29f28/service-front/web/src/javascript/sessionDialog.js) but with a few tweaks for our use case.

I encountered a few [limitations](https://github.com/jsdom/jsdom/issues/2102) with using jsdom in jest which is why the tests are split between unit and e2e.

Fixes MLPAB-981


web/assets/data-loss-warning.js
```diff
@@ -0,0 +1,130 @@
+export class DataLossWarning {
+    constructor(saveOrReturnComponent) {
```
hawx, line 2: I feel like we may as well just query for this in here since we are grabbing other stuff from `document`. And maybe use the tasklist `<a>` instead of the button group, because we don't do anything with the "save" bit

web/assets/data-loss-warning.js
```diff
@@ -0,0 +1,130 @@
...
+        const focusableEls = this.dialog.querySelectorAll('a[href], button')
+        const firstFocusableEl = focusableEls[0]
+        const lastFocusableEl = focusableEls[focusableEls.length - 1]
+        const KEY_CODE_TAB = 9
+        const KEY_CODE_ESC = 27
+
+        const tabPressed = (e.key === 'Tab' || e.keyCode === KEY_CODE_TAB)
+        const escPressed = (e.key === 'Esc' || e.keyCode === KEY_CODE_ESC)
+
+        if (!tabPressed && !escPressed) {
```
hawx, line 108: this is kind of pointless when we have explicit `if`s for these below

web/template/layout/data-loss-warning-dialog.gohtml
```diff
@@ -0,0 +1,18 @@
+{{ define "data-loss-warning-dialog" }}
+    <div id="dialog-overlay" class="govuk-!-display-none" tabindex="-1"></div>
```
hawx, line 2: you could incorporate the overlay in the modal so only one thing has to be toggled

web/assets/data-loss-warning.js
```diff
@@ -0,0 +1,130 @@
...
+            if (element.tagName === 'BUTTON') {
+                element.addEventListener('click', this.toggleDialogVisibility.bind(this))
+            }
+        }
+    }
+
+    handleTrapFocus(e) {
+        const focusableEls = this.dialog.querySelectorAll('a[href], button')
+        const firstFocusableEl = focusableEls[0]
+        const lastFocusableEl = focusableEls[focusableEls.length - 1]
```
hawx, line 101: I know this is copy-pasted, feel like since we know what the first and last are this could be simplified

web/assets/data-loss-warning.test.js
```diff
@@ -0,0 +1,200 @@
+import { DataLossWarning } from './data-loss-warning'
```
hawx, line 1: are we going to run these in GitHub? (I'd personally be fine with not having jest test and just covering the behaviour fully in cypress)

web/assets/data-loss-warning.js
acsauk: Yeah fair point - fell into the pattern of existing components we have.

web/assets/data-loss-warning.js
```diff
@@ -0,0 +1,130 @@
...
+
+        return this.buttonsPresent(2, this.saveOrReturnComponent)
+    }
+
+    dialogComponentValid() {
+        if (!this.dialog || !this.dialogOverlay) {
+            return false
+        }
+
+        return this.buttonsPresent(2, this.dialog.querySelector(".govuk-button-group"))
```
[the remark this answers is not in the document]
acsauk: My thinking was there are instances where we aren't using the `save-and-return-buttons` partial in a couple of places in the app, so we wouldn't want the warning to be applied in error if someone adds the partial warning component without having two buttons/being in the right context to use this.

Possibly overly defensive, it was based on future planning for people not familiar with the codebase. But I guess they will know if something is working/not working. OK, I've convinced myself while typing, I'll drop out!

web/assets/data-loss-warning.js
```diff
@@ -0,0 +1,130 @@
...
+                    if (this.changesMade()) {
+                        e.preventDefault()
+                    }
+                })
+
+                element.addEventListener('click', this.toggleDialogVisibility.bind(this))
+            }
+        }
+
+        for (let element of this.dialog.querySelector(".govuk-button-group").children) {
```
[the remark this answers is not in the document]
acsauk: Yep deffo 👍

web/assets/data-loss-warning.js
acsauk: Good point 👍

web/assets/data-loss-warning.test.js
acsauk: I was aiming to have all this in unit tests but the quirkiness of jsdom has just made it not worthwhile. I'll add a couple of lines to cypress to test the dialog is actually visible then I think we can safely drop this.

web/template/layout/data-loss-warning-dialog.gohtml
acsauk: Nice, hadn't thought of that. Will move it up.

web/template/layout/data-loss-warning-dialog.gohtml
```diff
@@ -0,0 +1,18 @@
...
+    <div id="dialog"
+         class="govuk-!-display-none govuk-!-padding-left-4 govuk-!-padding-top-2"
```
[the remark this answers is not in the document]
acsauk: I'd added this before I'd started applying the rest of the styling so good catch, I'll remove.

web/template/sign_your_lpa.gohtml
```diff
@@ -70,6 +70,8 @@
 
         {{ template "details" (details . "howTickingActsAsSignature" "howTickingActsAsSignatureContent" false) }}
 
+        {{ template "data-loss-warning-dialog" . }}
```
hawx, line 73: you'll also need the `id` on the task list link

hawx (approved):

PR #556, opened 2023-06-28
MLPAB-1097 Replace strings with string enums

hawx:


andrewpearce-digital (dismissed):

acsauk (approved):
Nice - this makes things much more readable around the codebase and in templates.

:shipit:

PR #573, opened 2023-07-11
MLPAB-1006, MLPAB-1007: Add different fee type page

acsauk:
# Purpose

Adds a page to ask the donor if they are applying for a different type of fee.

Fixes MLPAB-1006, MLPAB-1007

## Approach

I've dropped the content from about payment as its likely to be very different once this gets the content treatment. I've also moved `HappyForm` to live in `form` and renamed it to `YesNoForm` because that's what it is.


hawx (approved):

hawx (dismissed):
`form` is a much better place for it to live

hawx (approved):

PR #575, opened 2023-07-12
MLPAB-1013 & MLPAB-1128 Add pages for evidence required and whether can be uploaded

hawx:


app/internal/page/donor/can_evidence_be_uploaded.go
```diff
@@ -0,0 +1,41 @@
...
+
+func CanEvidenceBeUploaded(tmpl template.Template) Handler {
+	return func(appData page.AppData, w http.ResponseWriter, r *http.Request, lpa *page.Lpa) error {
+		data := &canEvidenceBeUploadedData{
+			App:     appData,
+			Options: actor.YesNoValues,
+		}
+
+		if r.Method == http.MethodPost {
+			form := form.ReadHappyForm(r)
```
acsauk, line 27: Ah cool - I think I'll replace my custom yes/no form for this one to keep in line 👍

acsauk (approved):
:shipit:

app/internal/page/donor/can_evidence_be_uploaded.go
hawx: it probably should be renamed to `YesNoForm` to make it obvious what it actually is

PR #578, opened 2023-07-12
MLPAB-1166 - fix path to s3 access logs

andrewpearce-digital:
# Purpose

This fixes access denied errors by using the correct bucket for access logging.


acsauk (approved):
:shipit:

PR #580, opened 2023-07-13
MLPAB-1019 Add page on how to send evidence that goes to payment

hawx:
I also extracted a little helper to wrap the payment logic because it is starting to get everywhere


app/internal/page/donor/register.go
```diff
@@ -409,3 +426,42 @@ func makeLpaHandle(mux *http.ServeMux, store sesh.Store, defaultOptions handleOp
...
+	sessionStore sessions.Store
+	payClient    PayClient
+	appPublicURL string
+	randomString func(int) string
+}
+
+func (p *payHelper) Pay(appData page.AppData, w http.ResponseWriter, r *http.Request, lpa *page.Lpa) error {
+	createPaymentBody := pay.CreatePaymentBody{
+		Amount:      page.CostOfLpaPence,
+		Reference:   p.randomString(12),
```
acsauk, line 441: This hasn't explicitly been said yet but based on the vega call earlier this week I think this will need to be the UID we get back from sending the basic donor details to Sirius to support the separate Sirius processes around matching payments.

acsauk (approved):
One comment on the pay reference but may be worth confirming with vega before we implement it (just wanted to get it out my head).

app/internal/page/donor/register.go
hawx: yeah, well at least now it is only specified in one place when we get round to that

PR #583, opened 2023-07-13
MLPAB-1163 - create an egress security group for lambda

andrewpearce-digital:
# Purpose

Prepare for a the av lambda to be able to egress the VPC

Fixes MLPAB-1163

## Approach

- create security group and rule permitting egress


acsauk (approved):
:shipit:

PR #584, opened 2023-07-13
MLPAB-1020: Send correct fee type cost to GOV UK Pay

acsauk:
# Purpose

Assigns a cost to `FeeTypes` and uses when making payment.

Fixes MLPAB-1020


andrewpearce-digital (approved):

app/internal/page/data.go
```diff
@@ -85,8 +83,21 @@ const (
 	HalfFee FeeType = iota + 1
 	NoFee
 	Hardship
+	Full
```
hawx, line 86: `FullFee` would make more sense when seen out of context

hawx: and since it is the default you could have it setup like:

```golang
//go:generate enumerator -type FeeType 
type FeeType uint8

const (
  FullFee FeeType = iota
  HalfFee
  NoFee
  Hardship // and maybe change this to HardshipFee too?
)
```

then you wouldn't need to set it in donorStore.Create

app/internal/page/data.go
```diff
@@ -85,8 +83,21 @@ const (
...
 )
 
+func (i *FeeType) Cost() int {
```
hawx, line 89: we'll be using this type directly so `(i FeeType)` as the receiver instead

app/internal/page/data.go
acsauk: Ah nice, I hadn't twigged that we get the default when instantiating a new Lpa. And yeah, agree on expanding the const naming.

app/internal/page/data.go
```diff
@@ -80,14 +80,14 @@ const (
 type FeeType uint8
 
 const (
-	HalfFee FeeType = iota + 1
+	FullFee FeeType = iota + 1
```
hawx, line 83: this needs to be `iota` so it starts off at 0 and works as the default, you also will want to change the generator comment not to have `-empty` (and `-trimprefix` isn't needed as the names aren't prefixed)

hawx: the underlying field is `uint8` which has the default value of `0`, so basically you just need to give the 0 value a name

hawx (approved):

PR #585, opened 2023-07-14
MLPAB-1014 Upload one file for evidence

hawx:
I'm going to guess at some point we'll want to upload multiple, but that will mostly be a design thing. Also we may accept more than PDFs, but I'm limiting it here just to show the idea. The 32Mb limit is very arbitrary too.


app/internal/app/app.go
```diff
@@ -160,7 +166,9 @@ func App(
 func withAppData(next http.Handler, localizer page.Localizer, lang localize.Lang, rumConfig page.RumConfig, staticHash, oneloginURL string) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		ctx := r.Context()
-		localizer.SetShowTranslationKeys(r.FormValue("showTranslationKeys") == "1")
+		if contentType, _, _ := strings.Cut(r.Header.Get("Content-Type"), ";"); contentType != "multipart/form-data" {
+			localizer.SetShowTranslationKeys(r.FormValue("showTranslationKeys") == "1")
```
hawx, line 170: we just need to avoid calling anything that parses the form if it is multipart

app/internal/page/donor/upload_evidence.go
```diff
@@ -0,0 +1,150 @@
...
+		return &uploadEvidenceForm{Error: err}
+	}
+
+	if part.FormName() != "upload" {
+		return &uploadEvidenceForm{Error: errors.New("unexpected field name")}
+	}
+
+	buf := bufio.NewReader(part)
+
+	sniff, _ := buf.Peek(peekSize)
```
acsauk, line 109: 🐽

app/internal/page/multi_read_closer.go
```diff
@@ -0,0 +1,99 @@
...
+		if len(mr.readClosers) == 1 {
+			if r, ok := mr.readClosers[0].(*multiReadCloser); ok {
+				mr.readClosers = r.readClosers
+				continue
+			}
+		}
+		n, err = mr.readClosers[0].Read(p)
+		if err == io.EOF {
+			// Use eofReader instead of nil to avoid nil panic
+			// after performing flatten (Issue 18232).
```
acsauk, line 39: Is this why we've bought in a modified version?

hawx: oh no, this could probably be cleaned up. The modification is to take something that handles `io.Reader`s (https://pkg.go.dev/io#MultiReader) to do `io.ReadCloser`s instead

acsauk (approved):
All looks good - one Q on bringing in the modified multi-reader.

PR #590, opened 2023-07-18
MLPAB-1026 Show additional states in task list for payment

hawx:


app/internal/actor/task_state.go
```diff
@@ -22,3 +22,14 @@ func (t TaskState) String() string {
 	}
 	return ""
 }
+
+//go:generate enumerator -type PaymentTask -trimprefix
+type PaymentTask uint8
```
hawx, line 27: In my mind the options were:

- extend `TaskState`, but then you'd need to know that these extras are only used for one task
- add another thing to `Lpa`, but then that would need to sync with the task state
- add another type, what I picked

the only downside is that it feels weird to have a custom field in the task list for it...

PR #592, opened 2023-07-18
MLPAB-1137 Use localizer to concatenate sentences

hawx:


lang/en.json
```diff
@@ -221,7 +222,6 @@
     "colleague": "Colleague",
     "healthProfessional": "GP or other qualified health professional",
     "legalProfessional": "Solicitor or other qualified legal professional",
-    "or": "Or",
```
hawx, line 224: I'm going to put a ticket in to remove unused translation strings, because I think we have quite a few now

acsauk (approved):
:shipit:

PR #595, opened 2023-07-19
MLPAB-1124: Update CP what happens next

acsauk:
# Purpose

Update what happens next content and add nav buttons. 

This kind of doesn't make sense until MLPAB-1121 is completed as the new journey logic moves the checks on if donor has signed and CP witnessed to task list links rather than our existing conditional redirects. 

Fixes MLPAB-1124


app/internal/page/fixtures.go
```diff
@@ -72,6 +72,10 @@ func Fixtures(tmpl template.Template) Handler {
 						values.Add("startCpFlowDonorHasNotPaid", "1")
 					}
 
+					if data.Form.Signed != "" {
+						values.Add("lpa.signedByDonor", "1")
```
hawx, line 76: are we saying that the LPA will always be signed for the flow to start now, so we could always set this?

acsauk: Not necessarily - from what I understand it will still be possible for a CP to start the flow while the donor has not signed. In this case the task list will have two sections rather than one and they'll be unable to get to the flow where they see the what happens next page and signing page. They'll have an option to view the draft LPA but all other section links will show as cannot start)

hawx (approved):

PR #605, opened 2023-07-25
MLPAB-1196 - Send uploaded documents for review

andrewpearce-digital:
# Purpose

When ready to submit, send uploaded documents in support of a reduced fees request to Sirius

Fixes MLPAB-1196

## Approach

This PR creates the resources to select objects to replicate using a S3 replication configuration. A further PR will set up the Batch Replication process using the replication configuration to create a manifest and select a replication IAM role

- configure object replication when tag criteria matches
- use replication role in account level
- ensure we're pulling the appropriate key for each environment
- disable replication in production 
- point to a stable non production environment

When an object is uploaded, it is scanned for malicious code, and safe objects are tagged `"virus-scan-status": "ok"`. When we are ready to replicate the objects, we should add another tag `"replicate": "true"` (we can change what this is if necessary).

A service role will carry out the object replication.

~Note: Need to check and prevent deletion of objects in the target bucket, when the object is removed from the source bucket.~

## Learning

- https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-how-setup.html
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_replication_configuration#filter
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication-config-for-kms-objects.html#replications


andrewpearce-digital:
image

hawx (approved):

PR #608, opened 2023-07-25
MLPAB-1149: Update /check-your-lpa and /lpa-details-saved variable template logic 

acsauk:
# Purpose

Now we contact the CP on the donor checking their LPA we can no longer rely on the paid task being completed to vary the content for these two pages. I've relied on a query param here as it seemed the most straightforward way of dealing with the first time the donor hits `/lpa-details-saved` when the check and send task will always be complete.

Fixes MLPAB-1149

### DIFF
--- a/app/internal/page/donor/check_your_lpa.go
+++ b/app/internal/page/donor/check_your_lpa.go
@@ -34,6 +34,12 @@ func CheckYourLpa(tmpl template.Template, donorStore DonorStore, shareCodeSender
 			data.Errors = data.Form.Validate()
 
 			if data.Errors.None() {
+				redirect := page.Paths.LpaDetailsSaved.Format(lpa.ID)
+
+				if !data.Completed {
+					redirect = redirect + "?firstCheck=1"
+				}
+
 				lpa.CheckedAndHappy = data.Form.CheckedAndHappy
 				lpa.Tasks.CheckYourLpa = actor.TaskCompleted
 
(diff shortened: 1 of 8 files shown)


web/template/lpa_details_saved.gohtml
```diff
@@ -36,7 +36,7 @@
             {{ if .Lpa.Tasks.PayForLpa.IsCompleted }}
                 <a href="{{ link .App (.App.Paths.Dashboard.String) }}" class="govuk-button">{{ tr .App "returnToDashboard" }}</a>
             {{ else }}
-                {{ if eq "paper" .Lpa.CertificateProvider.CarryOutBy }}
+                {{ if and (eq "paper" .Lpa.CertificateProvider.CarryOutBy) .IsFirstCheck }}
```
hawx, line 39: can't believe I missed something off when I did the enum stuff

app/internal/page/donor/check_your_lpa.go
```diff
@@ -34,6 +34,12 @@ func CheckYourLpa(tmpl template.Template, donorStore DonorStore, shareCodeSender
 			data.Errors = data.Form.Validate()
 
 			if data.Errors.None() {
+				redirect := page.Paths.LpaDetailsSaved.Format(lpa.ID)
```
hawx, line 37: this is an aside, but should we be auto-checking the `Form.CheckedAndHappy` with your previous answer now that it triggers new stuff?

hawx (approved):

app/internal/page/donor/check_your_lpa.go
acsauk: That's a good point - it feels like this should always require a reconfirmation especially when we come to implement Beth's new designs. I'll drop in the text message tickets I'm working on now.

web/template/lpa_details_saved.gohtml
acsauk: Doh, I was on autopilot. I'll bundle up with the next ticket too 👍

PR #609, opened 2023-07-26
MLPAB-940 Add pages to enter trust corporation details

hawx:


app/internal/actor/attorney.go
```diff
@@ -28,49 +28,101 @@ func (a Attorney) FullName() string {
...
+	// CompanyNumber as registered by Companies House
+	CompanyNumber string
+	// Email to contact the company
+	Email string
+	// Address of the company
+	Address place.Address
+}
+
+type Attorneys struct {
+	TrustCorporation TrustCorporation
```
hawx, line 44: The first commit shows how it was with `TrustCorporation`/`ReplacementTrustCorporation` on the `Lpa` type. I moved it here because I wanted to get the length of attorneys, and combining them with a `Len()` method seemed the most foolproof way

app/internal/actor/attorney.go
```diff
@@ -28,50 +28,112 @@ func (a Attorney) FullName() string {
...
+		}
+	}
+
+	return true
+}
+
+func (as Attorneys) Addresses() []place.Address {
+	var addresses []place.Address
+
+	if as.TrustCorporation.Address.String() != "" && !slices.Contains(addresses, as.TrustCorporation.Address) {
```
acsauk, line 73: Won't `addresses` always be an empty slice here?

acsauk (commented):
One Q and then good to go 👍

app/internal/actor/attorney.go
hawx: true true

acsauk (approved):
:shipit:

PR #610, opened 2023-07-26
MLPAB-1113, MLPAB-1114: Send SMS updates to paper and digital certificate providers

acsauk:
# Purpose

Adds checks on what stage the donor and CP are at in their journeys and sends appropriate SMSs.

Also enum-ed how CPs carry out their role.

Fixes MLPAB-1113, MLPAB-1114


acsauk:
Branch names with `+` is a no-go.

PR #612, opened 2023-07-28
MLPAB-1149: Move CP reference number after login to redirect returning CPs to dashboard

acsauk:
# Purpose

This change brings the CP journey in line with attorney journeys so the reference number only has to be entered once after login. This allows us to use the shared login and reference number handlers and ensures that returning CPs will be redirected to the dashboard when they return.

The dashboard store now returns certificate provider and attorney task lists to be used to vary content on the dashboard template and redirect to relevant parts of the flow.

Fixes MLPAB-1149


app/internal/app/dashboard_store.go
```diff
@@ -55,24 +64,65 @@ func (s *dashboardStore) GetAll(ctx context.Context) (donor, attorney, certifica
 		return nil, nil, nil, nil
 	}
 
-	var items []*page.Lpa
-	if err := s.dynamoClient.GetAllByKeys(ctx, searchKeys, &items); err != nil {
+	var lpasOrProvidedDetails []interface{}
+	if err := s.dynamoClient.GetAllByKeys(ctx, searchKeys, &lpasOrProvidedDetails); err != nil {
```
hawx, line 68: It will be better, now this is returning various types, to change the signature to `GetAllByKeys(context.Context, keys []Keys) ([]map[string]types.AttributeValue, error)`. Then you'll be able to inspect each list item for `PK` and `SK` to decide what it is, then use `attributevalue.UnmarshalMap` to get the full record

app/internal/page/enter_reference_number.go
```diff
@@ -50,20 +50,33 @@ func EnterReferenceNumber(tmpl template.Template, shareCodeStore ShareCodeStore,
...
 					LpaID:     shareCode.LpaID,
 				})
 
-				if _, err := attorneyStore.Create(ctx, shareCode.SessionID, shareCode.AttorneyID, shareCode.IsReplacementAttorney); err != nil {
-					var ccf *types.ConditionalCheckFailedException
-					if !errors.As(err, &ccf) {
-						return err
+				redirect := Paths.CertificateProvider.WhoIsEligible.Format(shareCode.LpaID)
+
+				if actorType.String() == actor.TypeCertificateProvider.String() {
```
hawx, line 60: these should be the same type, so you don't want the `.String()`s

app/internal/page/enter_reference_number.go
```diff
@@ -78,18 +91,18 @@ type enterReferenceNumberForm struct {
...
+		ReferenceNumber:    PostFormReferenceNumber(r, "reference-number"),
+		ReferenceNumberRaw: PostFormString(r, "reference-number"),
 	}
 }
 
 func (f *enterReferenceNumberForm) Validate() validation.List {
 	var errors validation.List
 
-	errors.String("reference-number", "twelveCharactersAttorneyReferenceNumber", f.ReferenceNumber,
+	errors.String("reference-number", "twelveCharactersReferenceNumber", f.ReferenceNumber,
```
hawx, line 102: probably shouldn't change these for the attorney page

app/internal/page/enter_reference_number.go
```diff
@@ -1,28 +1,28 @@
...
+	App    AppData
 	Errors validation.List
 	Form   *enterReferenceNumberForm
-	Lpa    *page.Lpa
+	Lpa    *Lpa
 }
 
-func EnterReferenceNumber(tmpl template.Template, shareCodeStore ShareCodeStore, sessionStore SessionStore, attorneyStore AttorneyStore) page.Handler {
-	return func(appData page.AppData, w http.ResponseWriter, r *http.Request) error {
+func EnterReferenceNumber(tmpl template.Template, shareCodeStore ShareCodeStore, sessionStore sessions.Store, certificateProviderStore CertificateProviderStore, attorneyStore AttorneyStore, actorType actor.Type) Handler {
```
hawx, line 24: it's kind of weird passing both stores in, I wouldn't mind this handler being duplicated for now as it is a bit custom for each

web/template/attorney_enter_reference_number.gohtml
```diff
@@ -10,9 +10,12 @@
       {{ trHtml .App "attorneyReferenceNumberContent" }}
 
       <form novalidate method="post">
-        {{ template "input" (input . "reference-number" "attorneyReferenceNumber" .Form.ReferenceNumberRaw "classes" "govuk-input--width-20 govuk-!-margin-bottom-3 govuk-input--extra-letter-spacing" "hint" "referenceNumberHint")}}
-        
-        {{ template "continue-button" . }}
+        <div class="govuk-form-group">
+          <fieldset class="govuk-fieldset">
```
hawx, line 14: we don't need a form-group or fieldset here as the `input` template handles it

web/template/certificate_provider_start.gohtml
```diff
@@ -8,7 +8,7 @@
       {{ trHtml .App "beACertificateProviderContent" }}
 
       <p class="govuk-body">
-        <a href="{{ link .App .App.Paths.CertificateProvider.EnterReferenceNumber.Format }}" role="button" draggable="false" class="govuk-button govuk-button--start" data-module="govuk-button">
+        <a href="{{ link .App .App.Paths.CertificateProvider.Login.String }}" role="button" draggable="false" class="govuk-button govuk-button--start" data-module="govuk-button">
```
hawx, line 11: can you use `.Format` in templates, it's a weak distinction for these paths but my thinking was that `.String` would only be used for registering and `.Format` would be used everywhere else, that way if their signature changes we'd get an error

app/internal/app/dashboard_store.go
acsauk: This is what I was trying to do in various iterations and failing - makes sense!

app/internal/page/enter_reference_number.go
acsauk: The previous keys for CP and Attorney ref content were a bit misleading:

```
"twelveCharactersAttorneyReferenceNumber": "your 12 character reference number"
"twelveCharactersReferenceNumber": "your 12 character certificate provider reference number"
```

Made a bit of a content decision here to just standardise both to be:

```
"twelveCharactersReferenceNumber": "your 12 character reference number",
```

app/internal/page/enter_reference_number.go
acsauk: I guess you're right - felt nice to be able to fully drop all the custom CP handler.

web/template/attorney_enter_reference_number.gohtml
acsauk: My bad ✅

web/template/certificate_provider_start.gohtml
acsauk: Sure - I wonder if there's any way to enforce this at the func level 🤔

app/internal/page/enter_reference_number.go
hawx: cool, I had assumed they were both specific

web/template/certificate_provider_start.gohtml
hawx: could write a little linter for the CI pipeline, but I think that would be overkill

app/internal/app/dashboard_store.go
```diff
@@ -27,7 +28,23 @@ type dashboardStore struct {
 	dynamoClient DynamoClient
 }
 
-func (s *dashboardStore) GetAll(ctx context.Context) (donor, attorney, certificateProvider []*page.Lpa, err error) {
+type keys struct {
+	PK, SK string
+}
+
+func (k keys) isLpa() bool {
+	return strings.Contains(k.SK, donorKey(""))
```
hawx, line 36: could use `strings.HasPrefix` for these to be more specific

app/internal/app/dashboard_store.go
```diff
@@ -55,24 +80,79 @@ func (s *dashboardStore) GetAll(ctx context.Context) (donor, attorney, certifica
...
+				donor = append(donor, page.LpaAndActorTasks{Lpa: lpa})
+			case actor.TypeAttorney:
+				attorneyMap[lpa.ID] = page.LpaAndActorTasks{Lpa: lpa}
+			case actor.TypeCertificateProvider:
+				certificateProviderMap[lpa.ID] = page.LpaAndActorTasks{Lpa: lpa}
+			}
+		}
+	}
+
+	for _, item := range lpasOrProvidedDetails {
```
hawx, line 114: it would be good to reorder the results in the test to have one of the provided details come before the LPA, just because I think it looks very "obvious" that you can join these loops when in fact you should not

acsauk: Mmmm good point - I tried a few different ways to condense it down into one loop but it was getting gross looking and hard to follow. I'll add an additional test to spell this out.

hawx (approved):

PR #631, opened 2023-08-09
MLPAB-1000 MLPAB-1002 Send event when previous application is being created

hawx:


app/internal/uid/client.go
```diff
@@ -53,9 +53,11 @@ type DonorDetails struct {
 }
 
 type CreateCaseRequestBody struct {
-	Type   string       `json:"type"`
-	Source string       `json:"source"`
-	Donor  DonorDetails `json:"donor"`
+	Type                      string       `json:"type"`
+	Source                    string       `json:"source"`
+	Donor                     DonorDetails `json:"donor"`
+	ApplicationReason         string       `json:"applicationReason"`
```
hawx, line 59: will be set to one of `new-application`, `remake` or `additional-application`

app/internal/uid/client.go
```diff
@@ -53,9 +53,11 @@ type DonorDetails struct {
...
+	PreviousApplicationNumber string       `json:"previousApplicationNumber,omitempty"`
```
hawx, line 60: will only be set if `applicationReason` is `remake` or `additional-application`

PR #655, opened 2023-08-24
MLPAB-1332 Send event for reduced fees instead of using dynamodb

hawx:


acsauk (approved):
:shipit: 

It's starting to get quite busy in `Put()` now so worth thinking of alternatives if there are future events to be emitted.

PR #656, opened 2023-08-24
MLPAB-1027: Move internal to root

acsauk:
# Purpose

Interim step so we can share code between app and lambda.

Fixes MLPAB-1027


hawx (approved):
this is so much better

PR #673, opened 2023-08-31
MLPAB-1352 - Fix notifications for path to live failures

andrewpearce-digital:
# Purpose

Path to live notifications when pipeline fails stopped working when the name of the path to live workflow changed.

Fixes MLPAB-1352

## Approach

- use correct name for path to live workflow


acsauk (approved):
:shipit:

PR #686, opened 2023-09-04
MLPAB-950 Add pages for when a donor cannot sign

hawx:


internal/actor/yes_no_maybe.go
```diff
@@ -0,0 +1,10 @@
+package actor
+
+//go:generate enumerator -type YesNoMaybe -linecomment -empty
+type YesNoMaybe uint8
+
+const (
+	Yes YesNoMaybe = iota + 1
+	No
+	Maybe
+)
```
hawx, line 10: kind of annoyed that I needed something with 3 states, and then couldn't think of a good way to differentiate from just yes/no

acsauk: I guess an alternative would be to use this around the app instead of `YesNo` but then the majority of pages wouldn't be using `Maybe` so yeah, necessary evil 🤷

PR #694, opened 2023-09-06
MLPAB-1030, MLPAB-1031: Handle fee denied event

acsauk:
# Purpose

Handles fee-denied events and ensures the amount to pay is correct.

I've replaced our modified open API spec for GOV UK Pay with the version from their repo now they've corrected the date format mistakes. To make the local version of the app work correctly I've had to add some named examples to the spec as we build the docker container that we reference in the mock server. This felt like a halfway house between keeping a local version of the spec and it getting out of date vs. taking the vanilla spec and then having the fee calculation act weirdly locally.

Fixes MLPAB-1030, MLPAB-1031


internal/templatefn/fn.go
```diff
@@ -375,3 +377,7 @@ func concatOr(app page.AppData, list []string) string {
 func concatComma(list []string) string {
 	return strings.Join(list, ", ")
 }
+
+func penceToPounds(pence int) string {
+	return humanize.CommafWithDigits(float64(pence)/100, 2)
```
acsauk, line 382: I was going a bit crazy trying to roll my own version of this but happy for any alternative on the PR 👍

hawx: I'm fine with this, but hope we'll never need commas

.codecov.yml
```diff
@@ -11,9 +11,9 @@ coverage:
     - "./internal/page/fixtures.go"
     - "./internal/page/testing_start.go"
     - "./cmd/mlpa/main.go"
-    - "./cmd/GOVUKNotify/main.go"
-    - "./cmd/GOVUKSignIn/main.go"
-    - "./cmd/OrdnanceSurveyPlacesAPI/main.go"
```
hawx, line 16: of course I'd miss somewhere

cmd/event-received/main.go
```diff
@@ -35,6 +35,10 @@ type moreEvidenceRequiredEvent struct {
 	UID string `json:"uid"`
 }
 
+type feeDeniedEvent struct {
+	UID string `json:"uid"`
+}
```
hawx, line 40: maybe we should have had one `uidEvent`

docker/mock-pay/Dockerfile
```diff
@@ -0,0 +1,10 @@
+
+FROM badouralix/curl-jq:latest as build
+
+RUN curl https://raw.githubusercontent.com/alphagov/pay-publicapi/master/openapi/publicapi_spec.json --output openapi-spec.json
+RUN cat openapi-spec.json | jq '.paths."/v1/payments/{paymentId}".get.responses."200".content."application/json".examples = { "full-fee": { value: { amount: 8200, payment_id: "hu20sqlact5260q2nanm0q8u93", reference: "your-reference" } }, "half-fee": { value: { amount: 4100, payment_id: "hu20sqlact5260q2nanm0q8u93", reference: "your-reference" } }}' > mktemp && mv mktemp openapi-spec.json
```
hawx, line 5: ha, this is a fun way of adding it

why `> mktemp && mv mktemp openapi-spec.json` and not `> openapi-spec.json` at the end? Or `> modified-openapi-spec.json` and then use that in the next step

internal/page/data.go
```diff
@@ -394,6 +394,20 @@ func (l *Lpa) TrustCorporationsNames() []string {
 	return names
 }
 
+func (l *Lpa) FeeAmount() int {
+	paid := 0
+
+	for _, payment := range l.PaymentDetails {
+		paid = paid + payment.Amount
```
hawx, line 401: += does exist in Go

acsauk: Haha I love that I can hear the tone in this comment 😂  noted ✅

cmd/event-received/main.go
acsauk: It is becoming quite apparent now...I was planning on holding off until the event catalogue is fully populated to see if there are any unexpected useful data points they are planning on sending over but seeing as everything working on just a UID makes sense to consolidate now 👍

docker/mock-pay/Dockerfile
acsauk: That works too - some leftover copy-paste from SO.

.codecov.yml
acsauk: 🤷

internal/templatefn/fn.go
acsauk: Things will have gone very wrong on the affordability front if so

hawx (approved):

PR #696, opened 2023-09-08
MLPAB-1339 Add page for signing on behalf of the donor

hawx:


internal/page/testing_start.go
```diff
@@ -247,6 +248,21 @@ func TestingStart(store sesh.Store, donorStore DonorStore, randomString func(int
...
+
+					lpa.AuthorisedSignatory = actor.AuthorisedSignatory{
+						FirstNames: "Alan",
+						LastName:   "Signatory",
+					}
+
+					lpa.IndependentWitness = actor.IndependentWitness{
+						FirstNames: "Iris",
+						LastName:   "Witness",
+					}
```
acsauk, line 263: There were some names suggested in the [UR doc](https://docs.google.com/document/d/1nT0pj_BZbVeDcG2TI13pX24wcJ6QZU7BfrTEa_NByOs/edit) recently for these two roles but I couldn't add them as they didn't exist in code - worth updating now to save future changes.

acsauk (commented):
Fixture name update suggestion and then good to go ✅

internal/page/testing_start.go
acsauk: Oh looks like that doc may be locked down. The names were:

Independent witness: Indie Irwin
Authorised signatory: Allie Adams

acsauk (approved):
:shipit:

PR #700, opened 2023-09-11
MLPAB-1340 Add flow for 2 witnesses

hawx:


internal/page/donor/witnessing_as_independent_witness_test.go
```diff
@@ -0,0 +1,395 @@
...
+		DonorIdentityUserData:   identity.UserData{OK: true, Provider: identity.OneLogin},
+		IndependentWitness:      actor.IndependentWitness{},
+		IndependentWitnessCodes: page.WitnessCodes{{Code: "1234", Created: now}},
+	})
+	resp := w.Result()
+
+	assert.Nil(t, err)
+	assert.Equal(t, http.StatusFound, resp.StatusCode)
+	assert.Equal(t, page.Paths.WitnessingAsCertificateProvider.Format("lpa-id"), resp.Header.Get("Location"))
+}
```
acsauk, line 148: Is there anything different being asserted here besides a witness being set on the LPA passed to the handler?

acsauk (approved):
One Q on a test but I may just be missing something subtle.

:shipit:

internal/page/donor/witnessing_as_independent_witness_test.go
hawx: you're right, it was some bad copypasta

acsauk (approved):
:shipit:

PR #732, opened 2023-09-27
MLPAB-1291 Hide submitted replacement attorney cards

hawx:


internal/page/attorney_fixtures.go
```diff
@@ -0,0 +1,264 @@
+package page
```
hawx, line 1: The fixtures/testing-start stuff has become too big. I think this approach of having a per-login actor handler will work better since it can make assumptions about the state of the data.

hawx: It's tempting to move it all to `internal/page/fixtures/...`

acsauk: Nice re-work - I'd be happy for this to live in a fixtures package at some point 👍

acsauk (approved):
👍

PR #744, opened 2023-10-03
MLPAB-1346: Ensure Puts only apply with the latest version of the struct

acsauk:
# Purpose

I initially started to look at being helpful and checking for a conditional fail when we try to PUT an out-of-date struct but that would have massively added to the complexity by trying to merge the two structs data. Given the edge case of this actually happening (e.g. an event updating an LPA between the start of a user submitting a form to update the same LPA) it felt like overengineering. In this solution, the user will get an error and can try again with the latest struct data. 

Fixes MLPAB-1346.


internal/dynamo/client.go
```diff
@@ -188,10 +189,27 @@ func (c *Client) Put(ctx context.Context, v interface{}) error {
...
+
+	if updatedAt, exists := item["UpdatedAt"]; exists {
+		var updatedAtTime time.Time
+		err = attributevalue.Unmarshal(updatedAt, &updatedAtTime)
+		if err != nil {
+			return err
+		}
+
+		if !updatedAtTime.IsZero() {
+			input.ConditionExpression = aws.String("UpdatedAt < :updatedAt")
```
hawx, line 205: Maybe it's too early for me to think, but I think this doesn't really change anything as it is "last write wins" which will still overwrite any other inflight writes

PR #746, opened 2023-10-04
MLPAB-1370 - create endpoints for logs, ssm, secrets manager and ecr

andrewpearce-digital:
# Purpose

Direct requests to AWS services via a VPC endpoint to reduce the egress vector through the internet gateway

Fixes MLPAB-1370

## Approach

- Create interface endpoints for services we use

## Learning

- https://docs.aws.amazon.com/vpc/latest/privatelink/privatelink-access-aws-services.html


acsauk (approved):
:shipit:

PR #747, opened 2023-10-04
Allow Sirius accounts to send events

gregtyler:
# Purpose

Add permissions to event bus to allow specified account to send events to the bus.

For VEGA-2090 #minor

## Approach

Re-used permissions set from Sirius. I added an environment-level config parameter called `receive_account_id`, if it's an empty string then the permission isn't created.

## Learning

VS Code was much more helpful with this than I expected.

## Checklist

* [ ] Changes to Github Actions jobs have been checked for all workflows

### DIFF
--- a/terraform/environment/region/modules/event_bus/main.tf
+++ b/terraform/environment/region/modules/event_bus/main.tf
@@ -53,3 +53,30 @@ resource "aws_cloudwatch_event_target" "cross_account_put" {
   role_arn       = var.iam_role.arn
   provider       = aws.region
 }
+
+# Allow other accounts to send messages
+data "aws_iam_policy_document" "main" {
+  statement {
+    sid    = "CrossAccountAccess"
+    effect = "Allow"
+    actions = [
+      "events:PutEvents",
+    ]
+    resources = [
+      aws_cloudwatch_event_bus.main.arn
+    ]
+
+    principals {
+      type        = "AWS"
+      identifiers = [var.receive_account_id]
+    }
+  }
+}
+
+resource "aws_cloudwatch_event_bus_policy" "main" {
+  count          = var.receive_account_id == "" ? 0 : 1
+  event_bus_name = aws_cloudwatch_event_bus.main.name
+  policy         = data.aws_iam_policy_document.main.json
+  provider       = aws.region
+}
+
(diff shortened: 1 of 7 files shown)


terraform/environment/region/variables.tf
```diff
@@ -90,3 +90,9 @@ variable "target_event_bus_arn" {
   type        = string
   description = "ARN of the event bus to forward events to"
 }
+
+variable "receive_account_id" {
+  type        = string
```
andrewpearce-digital, line 95: let's make these lists of strings

terraform/environment/region/modules/event_bus/main.tf
```diff
@@ -53,3 +53,30 @@ resource "aws_cloudwatch_event_target" "cross_account_put" {
...
+    actions = [
+      "events:PutEvents",
+    ]
+    resources = [
+      aws_cloudwatch_event_bus.main.arn
+    ]
+
+    principals {
+      type        = "AWS"
+      identifiers = [var.receive_account_id]
```
andrewpearce-digital, line 71: this can be just the var as it would be a list of strings now

andrewpearce-digital (changes requested):
let's make these lists of strings to allow us to grant access to services in other accounts

andrewpearce-digital (changes requested):
lets call the new policy and policy document `receive` as it's not the only policy in the module

andrewpearce-digital (approved):

terraform/environment/region/modules/event_bus/main.tf
```diff
@@ -53,3 +53,30 @@ resource "aws_cloudwatch_event_target" "cross_account_put" {
...
+      type        = "AWS"
+      identifiers = var.receive_account_ids
+    }
+  }
+}
+
+resource "aws_cloudwatch_event_bus_policy" "cross_account_receive" {
+  count          = length(var.receive_account_ids) > 0 ? 1 : 0
+  event_bus_name = aws_cloudwatch_event_bus.main.name
+  policy         = data.aws_iam_policy_document.macross_account_receivein.json
```
andrewpearce-digital, line 79:
```suggestion
  policy         = data.aws_iam_policy_document.cross_account_receive.json
```

andrewpearce-digital (approved):

PR #753, opened 2023-10-06
MLPAB-1471: Add file summary view and support more file types

acsauk:
# Purpose

I had to bring in `gabriel-vasile/mimetype` as the http content detection was struggling with some of the types appearing as zips and also had to bump the peek size to account for office document types.

Fixes MLPAB-1471


internal/page/donor/upload_evidence.go
```diff
@@ -51,29 +72,33 @@ func UploadEvidence(tmpl template.Template, payer Payer, donorStore DonorStore,
...
-						Body:                 bytes.NewReader(file.Data),
-						ServerSideEncryption: types.ServerSideEncryptionAwsKms,
-						Tagging:              aws.String("replicate=true"),
-					})
-					if err != nil {
-						return err
+				if form.Action == "upload" {
+					for _, file := range form.Files {
+						uuid := randomUUID()
+						key := lpa.UID + "-evidence-" + uuid
```
mattmachell, line 78: You might want to put a / rather than a - so that the bucket is easier to diagnose issues in when it's got a few more items in.

internal/page/donor/upload_evidence.go
```diff
@@ -51,29 +72,33 @@ func UploadEvidence(tmpl template.Template, payer Payer, donorStore DonorStore,
...
+							Key:                  aws.String(key),
+							Body:                 bytes.NewReader(file.Data),
+							ServerSideEncryption: types.ServerSideEncryptionAwsKms,
+							Tagging:              aws.String("replicate=true"),
+						})
+						if err != nil {
+							return err
+						}
+
+						lpa.EvidenceKeys = append(lpa.EvidenceKeys, page.Evidence{Key: key, Filename: file.Filename})
```
mattmachell, line 91: Is there any filename normalisation going on in `form.Files` that I've missed? Otherwise I can add some XSS in a filename to pass downstream.

internal/page/donor/upload_evidence.go
andrewpearce-digital: to add to this comment, / would be used to represent folder structures whereas - would appear as all objects in the root of a folder representation

https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-keys.html

hawx (approved):

PR #766, opened 2023-10-11
MLPAB-1094 Add summary cards for actors

hawx:


acsauk (approved):
This isn't anything to do with implementation but the donor card address is looking pretty smooshed up due to the mix of 4 and 3 columns in the card. Sure it will come up in design review when we look at this - we can discuss if we need the ID verified badge then.

Screenshot 2023-10-12 at 11 16 35

PR #779, opened 2023-10-17
MLPAB-1472: Update virus scanning results on evidence in dynamodb

acsauk:
# Purpose

Fixes MLPAB-1472


cmd/event-received/main.go
```diff
@@ -155,63 +177,118 @@ func handleFeeApproved(ctx context.Context, dynamoClient dynamodbClient, event e
...
 
-	if key.PK == "" {
-		return errors.New("PK missing from LPA in response to 'fee-denied'")
+	hasScannedTag := false
+	var tagIndex int
+
+	for i, tag := range tags {
+		if *tag.Key == "virus-scan-status" {
+			hasScannedTag = true
+			tagIndex = i
```
hawx, line 238: instead of storing the index to get later, would it be easier to do something like

```go
hasScannedTag := false
hasVirus := false
for _, tag := range tags {
  if *tag.Key == "virus-scan-status" {
    hasScannedTag = true
    hasVirus = tag.Value == virusFound
    break
  }
}
```

internal/page/data.go
```diff
@@ -182,49 +182,63 @@ type Lpa struct {
...
 	}
 
 	return keys
 }
 
-func (es *Evidences) GetByKey(key string) Evidence {
-	for _, e := range *es {
-		if e.Key == key {
-			return e
+func (e *Evidence) GetByDocumentKey(key string) Document {
```
hawx, line 216: this could just be `Get` as we aren't giving another way to retrieve elements

internal/page/data.go
```diff
@@ -182,49 +182,63 @@ type Lpa struct {
...
 
-	return Evidence{}
+	return Document{}
 }
 
-type Evidence struct {
-	Key      string
-	Filename string
-	Sent     time.Time
+func (e *Evidence) Update(document Document) bool {
```
hawx, line 226: for the others like Attorneys I changed these to be called `Put` and to always work as the `bool` return was never going to be useful, which I think is the same case here, so could match them?

acsauk: This makes more sense - I was trying to push a bit of the logic for the lambda in here but I'm already guarding against creating a new document there so should be fine.

hawx (approved):

PR #783, opened 2023-10-18
MLPAB-1472 - Persist scan results on files

andrewpearce-digital:
# Purpose

Show the user the status of their uploaded documents and if any contain malicious code or viruses

Fixes MLPAB-1472

## Approach

- run lambda when object is tagged 
- allow lambda to get an s3 object's tags

## Learning

-


acsauk (approved):
:shipit:

PR #785, opened 2023-10-19
MLPAB-1502 - Fix decrypt permissions for s3 antivirus scanning

andrewpearce-digital:
# Purpose

Prevent malicious code being forwarded from our service

Fixes MLPAB-1502

## Approach

- allow KMS decrypt using uploads bucket encryption key


acsauk (approved):
:shipit:

PR #788, opened 2023-10-20
MLPAB-1333 Allow setting identity confirmation in fixtures

hawx:


acsauk (approved):
:shipit: 

Not related to this PR but would be good to keep the fixtures nav links on the other actor pages too.

PR #796, opened 2023-10-24
MLPAB-1286 Separate registered LPAs on the dashboard

hawx:
I'm not sure how much putting it all in the template is cheating, but I kind of suspect this will get changed again...


web/template/dashboard.gohtml
```diff
@@ -5,10 +5,10 @@
 {{ define "donorCard" }}
   <div class="app-dashboard-card">
     <div class="app-dashboard-row">
-      <h3 class="govuk-heading-m">
+      {{ if .Item.Lpa.RegisteredAt.IsZero }}<h3 class="govuk-heading-m">{{ else }}<h4 class="govuk-heading-m">{{ end }}
```
hawx, line 8: how I wish XHTML 2.0 had won so we could just do `<h>`

acsauk: 😂 I am constantly parsing hierarchy on pages like this - nightmare!

acsauk (approved):
:shipit:

PR #799, opened 2023-10-24
MLPAB-1513 - Do not allow End of Workflow to be skipped

andrewpearce-digital:
# Purpose

If the End of Workflow job gets a status of skipped, it will count as successful and allow a PR to be merged

Fixes MLPAB-1513

## Approach

- Use success of the preceding important job
- Exit 0 if successful
- Exit 1 if not

## Learning

### DIFF
--- a/lang/cy.json
+++ b/lang/cy.json
@@ -1,5 +1,5 @@
 {
-    "cookiesOn": "Cwcis ar Gwneud atwrneiaeth arhosol'",
+    "cookiesOn": "Cwcis ar Gwneud atwrneiaeth arhosol",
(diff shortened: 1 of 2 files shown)


lang/cy.json
```diff
@@ -1,5 +1,5 @@
 {
-    "cookiesOn": "Cwcis ar Gwneud atwrneiaeth arhosol",
+    "cookiesOn": "Cwcis ar Gwneud atwrneiaeth arhosol'",
```
acsauk, line 2: Should this be here? (or should it be curly?)

andrewpearce-digital: It's a deliberate error to trip the unit test (the issue on the weblate PR)

Approval is to check the branch protection rather than merge

acsauk (dismissed):

hawx (approved):

PR #801, opened 2023-10-26
MLAPB-1514 - run replication job creation lambda every 2 mintues

andrewpearce-digital:
# Purpose

Move documents quickly when enabled

Fixes MLPAB-1514

## Approach

Set schedule to 2 minutes


acsauk (approved):
:shipit: speedy

PR #804, opened 2023-10-27
MLPAB-1306 Delete or withdraw an LPA

hawx:


internal/app/donor_store.go
```diff
@@ -198,6 +198,36 @@ func (s *donorStore) Put(ctx context.Context, lpa *page.Lpa) error {
...
+	}
+
+	keys, err := s.dynamoClient.AllKeysByPk(ctx, lpaKey(data.LpaID))
+	if err != nil {
+		return err
+	}
+
+	canDelete := false
+	for _, key := range keys {
+		if key.PK == lpaKey(data.LpaID) && key.SK == donorKey(data.SessionID) {
```
hawx, line 218: this is a bit defensive, but with it being a hard delete I think sensible

acsauk: Fair 👍

acsauk (approved):
:shipit:

PR #805, opened 2023-10-27
MLPAB-1473: Add holding modal while reduced fee evidence documents are scanned

acsauk:
# Purpose

Adds a holding modal that initiates a SSE connection to monitor when files are scanned.

I also had to add some optimistic locking to dynamo PUTs as the virus scan results were sometimes arriving within milliseconds creating a race condition that was overwriting data. ~~I've added 1 automatic retry to the lambda to try to mitigate this where we can but we'll need to see how this works to see if we need to rethink how we save the data (maybe to separate rows like the `evidence-received` event?)~~ I've moved what was called `Evidence` on the `Lpa` into `Documents` on individual rows to account for this.

I also found the spinner component in the [Home Office design system](https://design.homeoffice.gov.uk/components?name=Loading%20spinner) but it felt like overkill to install the whole design system for what is a few lines of CSS.

Fixes MLPAB-1473


internal/page/donor/upload_evidence.go
```diff
@@ -94,18 +99,47 @@ func UploadEvidence(tmpl template.Template, payer Payer, donorStore DonorStore,
...
+
+						return false
+					})
+
+					if len(filenames) > 0 {
+						if err := donorStore.Put(r.Context(), lpa); err != nil {
+							return err
+						}
+
+						// This is gross. There must be a nicer way to do this...
```
hawx, line 119: create an error type in `validation/error.go` to handle it, we should try to minimise the use of `CustomError`

internal/page/donor/witnessing_your_signature.go
```diff
@@ -24,6 +24,11 @@ func WitnessingYourSignature(tmpl template.Template, witnessCodeSender WitnessCo
 			if lpa.Donor.CanSign.IsYes() {
 				return appData.Redirect(w, r, lpa, page.Paths.WitnessingAsCertificateProvider.Format(lpa.ID))
 			} else {
+				lpa, err := donorStore.Get(r.Context())
```
hawx, line 27: why has this been added?

lang/en.json
```diff
@@ -840,5 +844,9 @@
     "lpaWithdrawn": "LPA withdrawn",
     "youHaveWithdrawnLpaNumber": "You have withdrawn LPA number <span class=\"govuk-!-font-weight-bold\">{{.UID}}</span>.",
     "opgWillNowContactAnyoneWhoHasAlreadyBeenContacted": "OPG will now contact anyone who has already been contacted in relation to your LPA and notify them of your decision to withdraw.",
-    "withdrawn": "Withdrawn"
+    "withdrawn": "Withdrawn",
+    "yourFilesAreUploading": "Your files are uploading",
```
hawx, line 848: image

the wording on this page is now weird, "Your files are uploading" even though they are in a section called "Uploaded files"

hawx:
I also think it would be nice if it were stored in separate rows, I think you'd be able to simplify the polling then instead of getting the Lpa each time

internal/page/donor/witnessing_your_signature.go
acsauk: Now we check the version of the Lpa on PUT we're getting a conditional expression error here as the PUTs happen so close together

lang/en.json
acsauk: Missed this - it will make more sense to have the modal appear when clicking upload files rather than continue.

internal/page/donor/witnessing_your_signature.go
hawx: ah, I forgot that it does the `Put` in the code sender

internal/app/app.go
```diff
@@ -49,13 +50,16 @@ type DynamoClient interface {
 	Put(ctx context.Context, v interface{}) error
 	Create(ctx context.Context, v interface{}) error
 	DeleteKeys(ctx context.Context, keys []dynamo.Key) error
+	DeleteOne(ctx context.Context, key dynamo.Key) error
```
hawx, line 53: maybe I should have made it `DeleteKeys(ctx context.Context, keys ...dynamo.Key)`

internal/app/app.go
```diff
@@ -49,13 +50,16 @@ type DynamoClient interface {
...
+	Update(ctx context.Context, input *dynamodb.UpdateItemInput) error
```
hawx, line 54: I'm not sure about exposing the `dynamodb` type here, we avoid it in the other signatures. Would `Update(ctx context.Context, pk, sk string, values map[string]types.AttributeValue, expression string) error` be any nicer?

internal/app/document_store.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/page"
+)
+
+type documentStore struct {
+	dynamoClient DynamoClient
+	s3Client     S3Client
+}
+
+func NewDocumentStore(dynamoClient DynamoClient, s3Client S3Client) DocumentStore {
+	return documentStore{dynamoClient: dynamoClient, s3Client: s3Client}
```
hawx, line 20: probably makes sense to have this return `&documentStore` and change all the receivers to `(s *documentStore)`

internal/app/document_store.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (s documentStore) GetAll(ctx context.Context) (page.Documents, error) {
+	data, err := page.SessionDataFromContext(ctx)
+	if err != nil {
+		return []page.Document{}, err
```
hawx, line 26: nothing major, but you should be able to return `nil` instead of `[]page.Document{}` for these

internal/app/document_store.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	if data.LpaID == "" {
+		return []page.Document{}, errors.New("documentStore.GetAll requires LpaID")
+	}
+
+	var ds []page.Document
+	if err := s.dynamoClient.AllByPartialSk(ctx, lpaKey(data.LpaID), DocumentKey(""), &ds); err != nil {
```
hawx, line 34: you could mash these ifs together to be 

```go
if err := s.dynamoClient.AllByPartialSk(ctx, lpaKey(data.LpaID), DocumentKey(""), &ds); err != nil && !errors.Is(err, dynamo.NotFoundError{}) {
  return nil, err
}
```

internal/app/document_store.go
```diff
@@ -0,0 +1,109 @@
...
+			return []page.Document{}, nil
+		}
+
+		return []page.Document{}, err
+	}
+
+	return ds, nil
+}
+
+func (s documentStore) UpdateScanResults(ctx context.Context, PK, SK string, virusDetected bool) error {
```
hawx, line 45: `pk, sk` here too

internal/app/document_store.go
```diff
@@ -0,0 +1,109 @@
...
+			})
+			s3Keys = append(s3Keys, d.Key)
+		}
+	}
+
+	if len(dynamoKeys) == 0 {
+		return nil
+	}
+
+	if err := s.s3Client.DeleteObjects(ctx, s3Keys); err != nil {
```
hawx, line 89: aren't we planning to manage the deletion in S3 of infected stuff with S3?

internal/app/document_store.go
```diff
@@ -0,0 +1,109 @@
...
+
+	return s.dynamoClient.DeleteKeys(ctx, dynamoKeys)
+}
+
+func (s documentStore) Delete(ctx context.Context, document page.Document) error {
+	if err := s.s3Client.DeleteObject(ctx, document.Key); err != nil {
+		return err
+	}
+
+	return s.dynamoClient.DeleteOne(ctx, dynamo.Key{
```
hawx, line 101: why do the `dynamo.Key` thing here, could just pass `pk, sk string` as it only gets decomposed

internal/app/document_store.go
```diff
@@ -0,0 +1,109 @@
...
+		PK: document.PK,
+		SK: document.SK,
+	})
+}
+
+func DocumentKey(s3Key string) string {
```
hawx, line 107: would prefer to keep this private

internal/app/donor_store.go
```diff
@@ -170,25 +179,35 @@ func (s *donorStore) Put(ctx context.Context, lpa *page.Lpa) error {
...
+				RequestType: lpa.FeeType.String(),
+				Evidence:    unsentKeys,
+			}); err != nil {
+				s.logger.Print(err)
+			} else {
+				for _, document := range documents {
+					if document.Sent.IsZero() && !document.Scanned {
+						document.Sent = s.now()
+						if err := s.documentStore.Put(ctx, document, nil); err != nil {
+							return err
```
hawx, line 208: We shouldn't be returning errors from here. It should just fail, log, and retry next time

hawx: also would be nice to batch these changes to documents up

internal/dynamo/client.go
```diff
@@ -232,24 +240,41 @@ func (c *Client) Put(ctx context.Context, v interface{}) error {
 		Item:      item,
 	}
 
-	if updatedAt, exists := item["UpdatedAt"]; exists {
-		var updatedAtTime time.Time
-		err = attributevalue.Unmarshal(updatedAt, &updatedAtTime)
+	// Tracking Value equality against data on write allows for optimistic locking
+	if currentVersion, exists := item["Version"]; exists {
```
hawx, line 244: do we still need this if documents are split out?

internal/page/data.go
```diff
@@ -177,56 +177,101 @@ type Lpa struct {
 	HasSentPreviousApplicationLinkedEvent bool
 }
 
-type Evidence struct {
-	Documents []Document
+type Document struct {
```
hawx, line 180: could all the document related stuff live in a new file

internal/page/donor/payment_confirmation.go
```diff
@@ -58,9 +58,14 @@ func PaymentConfirmation(logger Logger, tmpl template.Template, payClient PayCli
 		} else {
 			lpa.Tasks.PayForLpa = actor.PaymentTaskPending
 
-			for i, evidence := range lpa.Evidence.Documents {
-				if evidence.Sent.IsZero() {
-					err := evidenceS3Client.PutObjectTagging(r.Context(), evidence.Key, []types.Tag{
+			documents, err := documentStore.GetAll(r.Context())
```
hawx, line 61: do we need all of this since it happens in `payHelper.Pay`?

internal/page/donor/upload_evidence.go
```diff
@@ -74,46 +83,81 @@ func UploadEvidence(tmpl template.Template, payer Payer, donorStore DonorStore,
...
 
-						err := evidenceS3Client.PutObject(r.Context(), key, file.Data)
-						if err != nil {
+						document := page.Document{
+							PK:       "LPA#" + lpa.ID,
+							SK:       "#DOCUMENT#" + key,
+							Filename: file.Filename,
+							Key:      key,
+						}
+						if err := documentStore.Put(r.Context(), document, file.Data); err != nil {
```
hawx, line 97: it would be nice to provide a `documentStore.Create(ctx, lpa.ID, key, file.Filename)` method, then you won't have `LPA#` and `#DOCUMENT#` exposed here

cmd/event-received/main.go
```diff
@@ -256,26 +263,16 @@ func handleObjectTagsAdded(ctx context.Context, client dynamodbClient, event eve
...
 
-	document.Scanned = now()
-	document.VirusDetected = hasVirus
-
-	lpa.Evidence.Put(document)
-	lpa.UpdatedAt = now()
-
-	if err := client.Put(ctx, lpa); err != nil {
-		return fmt.Errorf("failed to update LPA for '%s': %w", objectTagsAddedEventName, err)
+	err = documentStore.UpdateScanResults(ctx, lpaKey.PK, "#DOCUMENT#"+objectKey, hasVirus)
```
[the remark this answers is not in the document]
acsauk: Much simpler - I was going around in circles with circular dependencies but makes more sense to hide the detail.

internal/app/app.go
acsauk: That works for me - slightly less boilerplate code to be passing in.

internal/app/document_store.go
acsauk: That was a discussion we'd had but nothing had been confirmed until today...I'll leave this here until we implement the feature incase we end up changing course.

internal/app/document_store.go
acsauk: Fell into the trap of copying the signature for `DeleteKeys()` and dropping the slice - will update 👍

internal/page/donor/payment_confirmation.go
acsauk: It does, but not for `HalfFee` - we wanted to be sure that the payment had been successful before tagging the docs as being ready for replication hence the replication.

hawx (approved):

PR #816, opened 2023-11-02
MLPAB-1497 Send full address in application-updated event

hawx:


scripts/get_event_schemas.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+
+mkdir -p internal/app/testdata
+curl -o internal/app/testdata/application-updated.json "https://raw.githubusercontent.com/ministryofjustice/opg-event-store/main/domains/POAS/events/application-updated/schema.json"
+curl -o internal/app/testdata/reduced-fee-requested.json "https://raw.githubusercontent.com/ministryofjustice/opg-event-store/main/domains/POAS/events/reduced-fee-requested/schema.json"
```
hawx, line 5: we could pull them each time the test is run, but I think having to manually update them will be less surprising

internal/app/donor_store_test.go
```diff
@@ -832,3 +837,52 @@ func TestDonorStoreDeleteWhenSessionMissing(t *testing.T) {
...
+			CreatedAt: time.Now(),
+			Donor: applicationUpdatedEventDonor{
+				FirstNames:  "syz",
+				LastName:    "syz",
+				DateOfBirth: date.New("2000", "01", "01"),
+				Address: place.Address{
+					Line1:      "line1",
+					Line2:      "line2",
+					Line3:      "line3",
+					TownOrCity: "townOrCity",
```
hawx, line 855: I wanted to be sure that the `json:` tag for this doesn't get changed without realising it will affect the event

scripts/get_event_schemas.sh
acsauk: This is cool for now - it may be worth thinking of some other way to enforce the schemas being up to date in some still undetermined way as we're only as safe as our brains remembering to run the script and commit.

acsauk (approved):
:shipit:

PR #818, opened 2023-11-02
MLPAB-1256, MLPAB-1254: Update hint text for jointly and severally mixed

acsauk:
# Purpose

Fixes MLPAB-1256, MLPAB-1254


web/template/how_should_attorneys_make_decisions.gohtml
```diff
@@ -25,7 +25,7 @@
                   {{ tr .App "jointlyAndSeverallyHumanised" }}
                 </label>
 
-                <div id="decision-type-item-hint" class="govuk-radios__hint">
+                <div id="decision-type-item-hint" class="govuk-hint govuk-radios__hint">
```
hawx, line 28: has someone asked off the ticket for this to be hint colour?

hawx (commented):
On the ticket they've put the last line as "You must state which decisions need to be agreed together.", which doesn't match what we currently have. But it isn't in green on the ticket. So should probably check what they want 

image

web/template/how_should_attorneys_make_decisions.gohtml
acsauk: Ah, my bad - I was grabbing the hint component from the design system and just assumed we'd missed the additional class. I'll remove.

hawx (approved):

PR #820, opened 2023-11-03
MLPAB-1422 Use an event to trigger getting a UID

hawx:
I've also:
- moved the event definitions to `event`
- made it so we do return the error if the event send fails, it's just going to make it much simpler to work with

I want to remove the need for those `HasSent...Event` fields at some point, but not right now


terraform/environment/region/modules/lambda/variables.tf
```diff
@@ -59,3 +59,8 @@ variable "ecr_arn" {
 variable "kms_key" {
   description = "KMS key for the lambda log group"
 }
+
+variable "iam_policy_document" {
+  type    = string
```
andrewpearce-digital, line 64: if we make this a list of strings type, with a default value of an empty list, we can utilise the merging feature inside the module. As it stands we can only pass in one policy document json.

terraform/environment/region/modules/lambda/iam.tf
```diff
@@ -49,7 +49,8 @@ data "aws_iam_policy_document" "lambda" {
     ]
   }
 
-  provider = aws.region
+  override_policy_documents = [var.iam_policy_document]
```
andrewpearce-digital, line 52: If we make the var type a list, we can omit the list and pass in just the var.

terraform/environment/region/modules/event_received/lambda.tf
```diff
@@ -7,20 +7,33 @@ module "event_received" {
...
-  ecr_arn     = var.lambda_function_image_ecr_arn
-  environment = data.aws_default_tags.current.tags.environment-name
-  kms_key     = data.aws_kms_alias.cloudwatch_application_logs_encryption.target_key_arn
-  timeout     = 300
-  memory      = 1024
+  image_uri           = "${var.lambda_function_image_ecr_url}:${var.lambda_function_image_tag}"
+  ecr_arn             = var.lambda_function_image_ecr_arn
+  environment         = data.aws_default_tags.current.tags.environment-name
+  kms_key             = data.aws_kms_alias.cloudwatch_application_logs_encryption.target_key_arn
+  iam_policy_document = data.aws_iam_policy_document.api_access_policy.json
```
andrewpearce-digital, line 16: I think this would become an iam policy document override variable that is a list if the other comments are accepted

terraform/environment/region/modules/lambda/variables.tf
```diff
@@ -59,3 +59,8 @@ variable "ecr_arn" {
 variable "kms_key" {
   description = "KMS key for the lambda log group"
 }
+
+variable "iam_policy_documents" {
+  type    = list(string)
```
andrewpearce-digital, line 64: because there's behaviour here, it would be worth adding a description, like "List of IAM policy documents that are merged together. Documents earlier in the list override later ones"

acsauk (approved):
:shipit:

PR #829, opened 2023-11-07
Rename `fee-denied` -> `reduced-fee-declined`

gregtyler:
To match [specification](https://ministryofjustice.github.io/opg-event-store/domains/POAS/events/reduced-fee-declined/).


hawx (dismissed):

acsauk (changes requested):
We'll need to update the event handler with the new name too - https://github.com/ministryofjustice/opg-modernising-lpa/blob/main/cmd/event-received/main.go#L166 (and note to self to see if we can add some testing around there to catch any event name mismatches)

gregtyler:
@acsauk Good catch, fixed!

acsauk (approved):
:shipit:

PR #832, opened 2023-11-07
MLPAB-1525, MLPAB-1521: Update certificate provider task flow

acsauk:
# Purpose

Fixes MLPAB-1525, MLPAB-1521


internal/actor/certificate_provider.go
```diff
@@ -22,6 +22,14 @@ const (
 	Online                                          // online
 )
 
+//go:generate enumerator -type CertificateProviderRelationshipLength -linecomment -trimprefix -empty
```
hawx, line 25: you don't need `-trimprefix` as the values aren't prefixed

internal/page/donor/how_do_you_know_your_certificate_provider.go
```diff
@@ -40,14 +40,12 @@ func HowDoYouKnowYourCertificateProvider(tmpl template.Template, donorStore Dono
...
 				}
 
-				if requireLength {
-					// TODO: should stay as Completed if editing and not changing the answer here
-					lpa.Tasks.CertificateProvider = actor.TaskInProgress
-				} else {
-					lpa.CertificateProvider.RelationshipLength = ""
-					lpa.Tasks.CertificateProvider = actor.TaskCompleted
+				if !requireLength {
+					lpa.CertificateProvider.RelationshipLength = 0
```
hawx, line 44: I think it would be nicer to have an explicit `RelationshipLengthUnknown` or whatever enum value, rather than magic 0

web/template/choose_address.gohtml
```diff
@@ -1,12 +1,28 @@
 {{ template "page" . }}
 
+
+
 {{ define "pageTitle" }}
+    {{ $manualTitle := "personsAddress" }}
```
hawx, line 6: thoughts on passing in a `Labels` struct or something to describe the labels that can be overridden for professional certificate providers vs. `.IsProfessionalCertificateProvider`?

hawx: I'm also noticing if I go back to the page the title isn't showing the "work" labels

hawx: image

and wondering if this should be listed as "Work address"?

internal/page/donor/how_do_you_know_your_certificate_provider.go
```diff
@@ -40,14 +40,12 @@ func HowDoYouKnowYourCertificateProvider(tmpl template.Template, donorStore Dono
...
 				}
 
+				lpa.Tasks.CertificateProvider = actor.TaskInProgress
+
```
hawx, line 48: Should we remove the address if they change how they know them? Otherwise we should only be making the task "in progress" here if they go pro->personal (as they have to answer how long they have known them)

internal/page/donor/how_do_you_know_your_certificate_provider.go
```diff
@@ -40,14 +40,12 @@ func HowDoYouKnowYourCertificateProvider(tmpl template.Template, donorStore Dono
...
-					// TODO: should stay as Completed if editing and not changing the answer here
-					lpa.Tasks.CertificateProvider = actor.TaskInProgress
-				} else {
-					lpa.CertificateProvider.RelationshipLength = ""
-					lpa.Tasks.CertificateProvider = actor.TaskCompleted
+				if !requireLength {
+					lpa.CertificateProvider.RelationshipLength = 0
 				}
 
+				lpa.Tasks.CertificateProvider = actor.TaskInProgress
```
hawx, line 47: I've just noticed we don't set this in `CertificateProviderDetails` which is where it should be

internal/actor/certificate_provider.go
acsauk: Agh sorry, copy pasta.

internal/page/donor/how_do_you_know_your_certificate_provider.go
acsauk: I hadn't thought about address but it feels right to start fresh if the relationship changes. And that makes sense, I think I'd oversimplified the existing logic when working out what should happen - I'll make it more explicit.

internal/page/donor/how_do_you_know_your_certificate_provider.go
acsauk: Good spot, I've added that in to the details handler now.

internal/page/donor/certificate_provider_address.go
```diff
@@ -12,19 +12,20 @@ import (
...
 func CertificateProviderAddress(logger Logger, tmpl template.Template, addressClient AddressClient, donorStore DonorStore) Handler {
 	return func(appData page.AppData, w http.ResponseWriter, r *http.Request, lpa *page.Lpa) error {
-		data := &chooseAddressData{
-			App:        appData,
-			ActorLabel: "certificateProvider",
-			FullName:   lpa.CertificateProvider.FullName(),
-			Form:       &form.AddressForm{},
+		data := newChooseAddressData(appData)
+		data.ActorLabel = "certificateProvider"
+		data.FullName = lpa.CertificateProvider.FullName()
```
hawx, line 17: would probably be nicer to have these as arguments to `newChooseAddressData`

hawx (approved):

PR #833, opened 2023-11-08
MLPAB-1579 - Expire objects that have been replicated

andrewpearce-digital:
# Purpose

Keep storage down and only hold documents as long as we need them

Fixes MLPAB-1579

## Approach

- add rule to expire objects with `replicate=true` after 30 days

## Learning

- https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html

### DIFF
--- a/terraform/environment/region/modules/uploads_s3_bucket/s3_bucket_lifecycles.tf
+++ b/terraform/environment/region/modules/uploads_s3_bucket/s3_bucket_lifecycles.tf
@@ -18,5 +18,22 @@ resource "aws_s3_bucket_lifecycle_configuration" "main" {
     status = "Enabled"
   }
 
+  rule {
+    id = "delete-objects-replicated-to-sirius"
+
+    filter {
+      tag {
+        key   = "replicate"
+        value = "true"
+      }
+    }
+
+    expiration {
+      days = "30"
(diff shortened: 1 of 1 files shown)


terraform/environment/region/modules/uploads_s3_bucket/s3_bucket_lifecycles.tf
```diff
@@ -18,5 +18,22 @@ resource "aws_s3_bucket_lifecycle_configuration" "main" {
     status = "Enabled"
   }
 
+  rule {
+    id = "delete-objects-replicated-to-sirius"
+
+    filter {
```
hawx, line 24: should we use a different filter, as the replicate tag is what we put on so doesn't mean it has been replicated?

andrewpearce-digital: We can only filter based on object prefix, tags or a combination.

We will be notified for each object that fails replication on the first attempt (currently within 15 minutes of the replicate tag being set).

Expiring objects after 30 days gives the chance to respond (manually at first automatically later if necessary).

hawx (approved):

PR #834, opened 2023-11-09
Add uids in all fixtures

hawx:


internal/page/fixtures/fixtures.go
```diff
@@ -140,3 +142,7 @@ func makePersonToNotify(name Name) actor.PersonToNotify {
 		},
 	}
 }
+
+func makeUid() string {
+	return strings.ToUpper("N-" + random.String(4) + "-" + random.String(4) + "-" + random.String(4))
```
hawx, line 147: it will look better being similar to reality, but I'm using an 'N' prefix so we know they are fake

acsauk (approved):
:shipit:

PR #836, opened 2023-11-10
Add tool to log events for use in cypress tests

hawx:


docker/localstack/localstack-init.sh
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-set -e
```
hawx, line 3: Sometimes localstack will fail to create things if they already exist, which breaks things that get setup after, which is very annoying. So I'm removing this to get the whole script to execute

internal/event/events.go
```diff
@@ -37,6 +37,6 @@ type PreviousApplicationLinked struct {
 type ReducedFeeRequested struct {
 	UID              string   `json:"uid"`
 	RequestType      string   `json:"requestType"`
-	Evidence         []string `json:"evidence"`
+	Evidence         []string `json:"evidence,omitempty"`
```
hawx, line 40: https://github.com/ministryofjustice/opg-event-store/pull/15 when I wrote the Cypress test we were sending `null` here, which didn't match the spec either. So, first bug found by the new tool!

cmd/event-logger/main.go
```diff
@@ -0,0 +1,166 @@
...
+)
+
+type message struct {
+	Time       time.Time
+	Detail     json.RawMessage
+	DetailType string `json:"detail-type"`
+	Source     string
+}
+
+const messageCount = 10
```
acsauk, line 31: I had to parse around the code below a bit to work out the purpose of this - would `messageLimit` or `maxMessages` be any clearer on intent?

docker/localstack/localstack-init.sh
acsauk: I didn't realise it was this opt that was getting in the way - no more `down` then `up` 🙌

acsauk (approved):
This is so cool - feels much safer now to be able to build events into our feature tests.

acsauk (approved):
:shipit:

PR #850, opened 2023-11-16
MLPAB-1365 Prevent re-checking LPA when no changes have been made

hawx:


internal/page/data.go
```diff
@@ -70,14 +71,16 @@ const (
 // Lpa contains all the data related to the LPA application
 type Lpa struct {
 	PK, SK string
+	// Hash is used to determine whether the Lpa has been changed since last read
+	Hash uint64 `hash:"-"`
```
hawx, line 75: I'm trying to not ignore too many fields, only those that contain hash values or are updated in `donorStore.Put` after the hash field is set

internal/page/donor/check_your_lpa.go
```diff
@@ -13,22 +15,102 @@ import (
...
+	App         page.AppData
+	Errors      validation.List
+	Lpa         *page.Lpa
+	Form        *checkYourLpaForm
+	Completed   bool
+	CanContinue bool
 }
 
-func CheckYourLpa(tmpl template.Template, donorStore DonorStore, shareCodeSender ShareCodeSender, notifyClient NotifyClient, certificateProviderStore CertificateProviderStore) Handler {
+type checkYourLpaNotifier struct {
```
hawx, line 26: I pulled this thing out because the code here was getting very nested

internal/page/data.go
```diff
@@ -124,6 +125,10 @@ type Lpa struct {
 	WantToApplyForLpa bool
 	// Confirmation that the applicant wants to sign the LPA
 	WantToSignLpa bool
+	// CheckedAt is when the donor checked their LPA
+	CheckedAt time.Time
```
hawx, line 129: I initially changed this because I was going to use it for the compare, but it would have ended up more complicated. I can put it back to a `bool` but think this is better anyway?

internal/page/data_test.go
```diff
@@ -147,6 +147,18 @@ func TestReplacementAttorneysStepIn(t *testing.T) {
...
+func TestGenerateHash(t *testing.T) {
+	lpa := &Lpa{}
+	hash, err := lpa.GenerateHash()
+	assert.Nil(t, err)
+	assert.Equal(t, uint64(6045214389284053903), hash)
+
+	lpa.ID = "1"
+	hash, err = lpa.GenerateHash()
+	assert.Nil(t, err)
+	assert.Equal(t, uint64(10213271685641464943), hash)
```
hawx, line 159: these numbers will change when we add/remove fields from Lpa, but I think that is ok because at least we'll know it is working

internal/page/donor/check_your_lpa.go
acsauk: Much easier to parse now

internal/page/data.go
acsauk: Fine with me - leaves things open for future requests on when things have happened.

internal/app/donor_store.go
```diff
@@ -130,6 +135,13 @@ func (s *donorStore) Latest(ctx context.Context) (*page.Lpa, error) {
 }
 
 func (s *donorStore) Put(ctx context.Context, lpa *page.Lpa) error {
+	newHash, err := lpa.GenerateHash()
+	if newHash == lpa.Hash || err != nil {
+		return err
```
acsauk, line 140: For instances where `Lpa` has not changed should we not do anything rather than return an error? I can see you have a test for that but I'm not getting how it's passing if we return an error here for a matching hash?

acsauk (commented):
One question on unchanged LPAs 🤔

internal/app/donor_store.go
hawx: `err` will be `nil` if `GenerateHash` worked fine, this is just me combining two ifs, like:

```go
if err != nil {
  return err
}
if newHash == lpa.Hash {
  return nil // but also err==nil as the if above didn't return
}
```

acsauk (approved):
:shipit:

PR #853, opened 2023-11-16
Refactor redirects

hawx:
Moves the redirect call to the path, so each type of path can require the parameters it needs


acsauk (approved):
Tidy - feels more natural this being on `Path`s.

:shipit:

PR #854, opened 2023-11-17
Move types related to Lpa to actor

hawx:
`actor` hasn't been a good name for this package for a while, but renaming it will be really annoying... and thinking of a good name is hard. Either way, the `Lpa` type does need to move from `page` to unlock other refactorings. This is the prep work of moving the types it references


acsauk (approved):
:shipit: 

I'm sure the appropriate name will make itself known to us once the dust has settled.

PR #856, opened 2023-11-17
MLPAB-1277: Update certificate provided content and format print media view

acsauk:
# Purpose

Fixes MLPAB-1277.

Print view looks like:

Screenshot 2023-11-20 at 16 33 31


web/template/layout/login-header.gohtml
```diff
@@ -92,7 +92,7 @@
         <div class="govuk-grid-column-two-thirds">
           <h2 class="service-header__heading">{{ tr .App "serviceName" }}</h2>
         </div>
-        <div class="govuk-grid-column-one-third">
+        <div class="govuk-grid-column-one-third app-hide-print">
```
hawx, line 95: it might look better hiding the top-level govuk header too, so we have the title of the service at the top of the page?

hawx (approved):
are you doing the email separately?

web/template/layout/login-header.gohtml
acsauk: I was on the fence with this - having govuk there gives it a bit of official-ness but agree the onelogin stuff isn't needed. I'm happy to remove for now and we can see what people think in testing.

hawx: The thing is we don't want people to think this is official or can be used as proof of something. It is just for record keeping so someone can remember they did a thing at a time

acsauk: Excellent point, I'd forgotten about that.

hawx (approved):

PR #857, opened 2023-11-17
MLPAB-1634 Set when lpa can be used if lpa type set to hw

hawx:
I felt like this would be simpler than having to check when we display it


acsauk (approved):
:shipit:

PR #864, opened 2023-11-21
MLPAB-1636: Update UID store UID reference

acsauk:
# Purpose

I should have tested this step locally on the last PR - all looks to be hooked up now.

Fixes MLPAB-1636


hawx (dismissed):
lesson is, never rename these

hawx (approved):

PR #865, opened 2023-11-21
MLPAB-1606 Add other versions of yes/no used in Welsh

hawx:


acsauk (approved):
Cool approach inverting the translation key :shipit:

PR #872, opened 2023-11-22
MLPAB-1583: Update paper certificate provider meeting prompt SMS template ID

acsauk:
# Purpose

Only this template had changes to the personalisation so only one new template required


internal/page/app_data.go
```diff
@@ -26,6 +26,7 @@ type AppData struct {
 	ActorType        actor.Type
 	AttorneyID       string
 	OneloginURL      string
+	AppPublicURL     string
```
acsauk, line 29: This felt like a natural place for this to live and have handy. I've not extended to packages other than `donor` yet but can do so in a follow-up PR if we're happy with it.

hawx (approved):

PR #874, opened 2023-11-23
MLPAB-1534 Remove certificate provider if select related

hawx:


web/assets/scss/main.scss
```diff
@@ -53,21 +53,22 @@ body:not(.js-enabled) .govuk-back-link {
 }
 
 .app-button-link {
```
hawx, line 55: for some reason I just forgot we had this style, I've added `display: contents` so it wraps in text like a link would

acsauk (approved):
:shipit:

PR #885, opened 2023-11-28
MLPAB-1598 Deploy mock-onelogin to ecs

hawx:


terraform/environment/region/variables.tf
```diff
@@ -37,6 +37,16 @@ variable "app_service_container_version" {
   description = "(optional) describe your variable"
 }
 
+variable "mock_onelogin_service_repository_url" {
+  type        = string
+  description = "(optional) describe your variable"
```
andrewpearce-digital, line 42: These variables aren't optional if they don't have a default value

terraform/environment/region/dns.tf
```diff
@@ -29,3 +30,27 @@ resource "aws_route53_record" "app" {
     create_before_destroy = true
   }
 }
+
+resource "aws_route53_record" "mock_onelogin" {
+  # *-mock-onelogin.app.modernising.opg.service.justice.gov.uk
+  count          = length(module.mock_onelogin)
```
andrewpearce-digital, line 36: This didn't need to change, it's still `var.mock_onelogin_enabled ? 1 : 0`

andrewpearce-digital (approved):

PR #892, opened 2023-11-30
MLPAB-1571 - Run replication job every 20 minutes

andrewpearce-digital:
# Purpose

This is to reduce the cost of S3 batch jobs

Fixes MLPAB-1571

## Approach

- Set schedule that triggers lambda to run every 20 minutes


acsauk (approved):
:shipit:

PR #894, opened 2023-12-01
MLPAB-1489: Add contact language preferences for all actors

acsauk:
# Purpose

Fixes MLPAB-1489


internal/localize/lang.go
```diff
@@ -1,26 +1,17 @@
...
+//go:generate enumerator -type Lang -linecomment -empty
+type Lang byte
 
-func (l Lang) String() string {
-	if l == Cy {
-		return welshAbbreviation
-	}
-
-	return englishAbbreviation
-}
```
acsauk, line 11: I think we were relying on this in summary pages but I'm not sure it makes sense now its an explicit choice the user has to make.

internal/page/donor/your_preferred_language.go
```diff
@@ -0,0 +1,49 @@
...
+
+func YourPreferredLanguage(tmpl template.Template, donorStore DonorStore) Handler {
+	return func(appData page.AppData, w http.ResponseWriter, r *http.Request, donor *actor.DonorProvidedDetails) error {
+		data := &yourPreferredLanguageData{
+			App: appData,
+			Form: &form.LanguagePreferenceForm{
+				Preference: donor.ContactLanguagePreference,
+			},
+			Options:    localize.LangValues,
+			FieldNames: form.FieldNames,
```
hawx, line 30: maybe it would be better to have `FieldName` and just pass in the one needed, if we carry on this pattern

web/template/layout/donor-details.gohtml
```diff
@@ -47,6 +47,20 @@
             </dd>
           {{ end }}
         </div>
+
+        {{ if $.App.IsDonor }}
+          <div class="govuk-summary-list__row">
+            <dt class="govuk-summary-list__key">{{ tr .App "preferredContactLanguage" }}</dt>
```
hawx, line 53: image

hawx (approved):

PR #902, opened 2023-12-05
MLPAB-1596 Send attorney invite after certificate is provided

hawx:
I've been wanting to make the notify personalisation stronger, so this adds a new `SendEmail` method which take structs for the emails. If you like it, I can change the others to it later, otherwise I can revert it


.codecov.yml
```diff
@@ -5,16 +5,17 @@ coverage:
...
-    - "./internal/telemetry"
-    - "./internal/page/fixtures"
     - "./cmd/event-logger/main.go"
     - "./cmd/event-received/main.go"
     - "./cmd/mlpa/main.go"
     - "./cmd/mock-notify/main.go"
     - "./cmd/mock-onelogin/main.go"
     - "./cmd/mock-os-api/main.go"
+    - "./internal/identity/yoti*"
+    - "./internal/notify/emails.go"
```
hawx, line 15: there's no value in testing this file, but that means we have to keep this file to only have things like I added here

acsauk (approved):
:shipit:

PR #909, opened 2023-12-08
MLPAB-1694 Retrieve openid configuration periodically

hawx:


andrewpearce-digital (approved):

internal/onelogin/configuration.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	return c.currentConfiguration.TokenEndpoint, c.currentJwks.Keyfunc, c.currentConfiguration.Issuer, nil
+}
+
+// requestRefresh will request that the configuration is refreshed, if no other request is waiting
+func (c *configurationClient) requestRefresh() {
+	select {
+	case c.refreshRequest <- struct{}{}:
+	default:
```
acsauk, line 104: I've just read up on `select` but I'm still a bit hazy on the need behind queueing up refresh requests 🤔

internal/onelogin/configuration_test.go
```diff
@@ -0,0 +1,243 @@
...
+	assert.Equal(t, expectedConfiguration, client.currentConfiguration)
+}
+
+func TestConfigurationClientBackgroundRefreshCancel(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	client := &configurationClient{ctx: ctx, now: time.Now}
+
+	cancel()
+	client.backgroundRefresh()
```
acsauk, line 207: Is this asserting anything? Or is it implicitly passing by the fact it doesn't throw an error?

acsauk (approved):
:shipit:

internal/onelogin/configuration_test.go
hawx: It is a bit of a rubbish test, I just wanted something to show that the `cancel()` call does stop it looping. So this test will hang if you remove that line

internal/onelogin/configuration.go
hawx: I stole this idea from the jwks package, it's just to prevent having multiple refresh requests piling up if one is already requested

PR #913, opened 2023-12-11
MLPAB-1278: Redirect to dashboard when OneLogin sub is linked to an LPA

acsauk:
# Purpose

I've added an extra field to the fixture pages to support logging back into the app by LPA UID. I was going to drop it after I'd worked out some bugs during manual testing but this felt like a missing feature anyway for any scenarios where you want to hop between actors on an LPA quickly. 

Fixes MLPAB-1278


Makefile
```diff
@@ -85,18 +85,18 @@ run-structurizr-export:
 	export -workspace /usr/local/structurizr/workspace.dsl -format mermaid
 
 scan-lpas: ##@app dumps all entries in the lpas dynamodb table
-	docker compose -f docker/docker-compose.yml exec localstack awslocal dynamodb scan --table-name lpas
+	docker compose -f docker/docker-compose.yml exec localstack awslocal dynamodb --region eu-west-1 scan --table-name lpas
```
acsauk, line 88: These commands weren't working for me after we bumped the localstack version. I played around with the shared DB env var but I was seeing some weirdness happening so this felt like a safer/less annoying option.

hawx: ah yeah, I hadn't used these since but it is what I had to do to the localstack init

cmd/mock-onelogin/main.go
```diff
@@ -157,6 +163,11 @@ func authorize() http.HandlerFunc {
 			user:     r.FormValue("user"),
 			identity: wantsIdentity,
 		}
+
+		if sub := r.FormValue("sub"); sub != "" {
+			loginWithSub[code] = sub
```
hawx, line 168: could this be added to `sessionData` instead?

hawx: I don't really like the mock routes doing things outside of the actual API. A better solution might be a separate route to do the setup, or provide a UI to input the option (like the identity bit does)?

acsauk: The UI option makes more sense than entering the UID on the fixtures page as I was struggling to give a route to test this from the start page - I'll add this in 👍

web/template/attorney_fixtures.gohtml
```diff
@@ -6,6 +6,11 @@
   <p class="govuk-body">Entering an email will cause a reference code to be sent and redirect you to the start page. Leave it blank to be signed in and taken to the dashboard instead.</p>
   {{ template "input" (input . "email" "email" "" "classes" "govuk-input--width-20" "type" "email" "spellcheck" "false" "autocomplete" "email") }}
 
+  {{ template "input" (input . "attorneySub" "Attorney OneLogin sub" .Sub "classes" "govuk-input--width-20") }}
```
hawx, line 9: you're doing `r.FormValue("sub")` in the code for each of these, so these aren't being picked up

acsauk: 🤦 good spot

hawx (approved):

hawx (changes requested):
actually, I'm not sure if this doesn't go to the reference page now? or is it just broken locally?

hawx (approved):
rebuilt containers and it worked 🤷

acsauk:
Same 🤷

PR #914, opened 2023-12-12
MLPAB-1637 Send donor lpa details to lpa store on submission

hawx:


internal/page/fixtures/fixtures.go
```diff
@@ -138,10 +142,11 @@ func makePersonToNotify(name Name) actor.PersonToNotify {
...
 			TownOrCity: "BIRMINGHAM",
 			Postcode:   "B14 7ED",
+			Country:    "GB",
 		},
 	}
 }
 
 func makeUID() string {
-	return strings.ToUpper("N-" + random.String(4) + "-" + random.String(4) + "-" + random.String(4))
+	return strings.ToUpper("M-" + random.String(4) + "-" + random.String(4) + "-" + random.String(4))
```
hawx, line 151: having to change this or the openapi spec will not verify the request

andrewpearce-digital (approved):

internal/lambda/client.go
```diff
@@ -0,0 +1,72 @@
...
+	SignHTTP(context.Context, aws.Credentials, *http.Request, string, string, string, time.Time, ...func(options *v4.SignerOptions)) error
+}
+
+//go:generate mockery --testonly --inpackage --name Doer --structname mockDoer
+type Doer interface {
+	Do(*http.Request) (*http.Response, error)
+}
+
+// A Client makes HTTP requests to AWS Lambda functions
+type Client struct {
```
acsauk, line 29: Should we be exporting this if we have a `New()`?

internal/lpastore/client_test.go
```diff
@@ -0,0 +1,400 @@
...
+		PactFileWriteMode: "merge",
+		LogDir:            "../../logs",
+		PactDir:           "../../pacts",
+	}
+	defer pact.Teardown()
+
+	pact.
+		AddInteraction().
+		Given("The lpa store is available").
+		UponReceiving("A request for a new case").
```
acsauk, line 233: Should we qualify the expected response by using something like:

```suggestion
		UponReceiving("A request for a new case using an existing UID").
```

Or does that not really matter? Its been a while since I used Pact so have forgotten most of the do's and don'ts.

internal/page/donor/register_test.go
```diff
@@ -23,7 +23,7 @@ import (
 
 func TestRegister(t *testing.T) {
 	mux := http.NewServeMux()
-	Register(mux, &log.Logger{}, template.Templates{}, nil, nil, &onelogin.Client{}, &place.Client{}, "http://example.org", &pay.Client{}, nil, &mockWitnessCodeSender{}, nil, nil, nil, nil, &notify.Client{}, nil, nil, nil, &mockDashboardStore{})
+	Register(mux, &log.Logger{}, template.Templates{}, nil, nil, &onelogin.Client{}, &place.Client{}, "http://example.org", &pay.Client{}, nil, &mockWitnessCodeSender{}, nil, nil, nil, nil, &notify.Client{}, nil, nil, nil, nil, nil)
```
acsauk, line 26: I'd started to use the mock objects here as its hard to track which nil corresponds to which argument...probably should have committed fully and updated them all but are you opposed to using mocks here instead of nils?

terraform/environment/terraform.tfvars.json
```diff
@@ -334,6 +379,15 @@
...
       },
+      "lpa_store_service": {
+        "base_url": "https://lpa-store.api.opg.service.justice.gov.uk",
+        "api_arns": [
+          "arn:aws:execute-api:eu-west-1:764856231715:*/*/PUT/lpas/*",
+          "arn:aws:execute-api:eu-west-2:764856231715:*/*/PUT/lpas/*",
+          "arn:aws:execute-api:eu-west-1:764856231715:*/*/GET/health-check",
+          "arn:aws:execute-api:eu-west-2:764856231715:*/*/GET/health-check"
+        ]
+      },
```
acsauk, line 390: Assuming this is fine but thought I should check, is the LPA store being used in any live environments yet?

acsauk (commented):
A couple of suggestions and questions and then ready to merge.

internal/lpastore/client_test.go
hawx: I will be adding more of these when I add the other data, so will expand them then

internal/page/donor/register_test.go
hawx: oh right, fine with me, I'll not change them again then

terraform/environment/terraform.tfvars.json
hawx: it is deployed to production, it doesn't do anything though

PR #915, opened 2023-12-12
MLPAB-1684 - make sure to tag without the main prefix too

andrewpearce-digital:
# Purpose

Fix missing tag on path to live run

Fixes MLPAB-1684

## Approach

- Tag image


acsauk (approved):
:shipit:

PR #922, opened 2023-12-13
MLPAB-1663 - Continuously monitor dependency health checks

andrewpearce-digital:
# Purpose

Let the team know when our service cannot reach it's dependencies

Fixes MLPAB-1663

## Approach

- create dependency alarms
- forward service alarms to pagerduty
- forward dependency alarms to pagerduty
- TODO: decide which environments should have alarm actions enabled

## Learning

### DIFF
--- a/terraform/environment/region/ecs.tf
+++ b/terraform/environment/region/ecs.tf
@@ -53,7 +53,7 @@ module "app" {
   aws_rum_guest_role_arn                               = data.aws_iam_role.rum_monitor_unauthenticated.arn
   rum_monitor_application_id_secretsmanager_secret_arn = aws_secretsmanager_secret.rum_monitor_application_id.id
   uid_base_url                                         = var.uid_service.base_url
-  lpa_store_base_url = var.lpa_store_service.base_url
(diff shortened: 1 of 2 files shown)


hawx (approved):

terraform/environment/region/health_checks.tf
```diff
@@ -0,0 +1,142 @@
...
+  sqs_success_feedback_sample_rate         = 100
+  provider                                 = aws.global
+}
+
+resource "aws_route53_health_check" "service_health_check" {
+  fqdn              = aws_route53_record.app.fqdn
+  reference_name    = "${substr(data.aws_default_tags.current.tags.environment-name, 0, 20)}-service-hc"
+  port              = 443
+  type              = "HTTPS"
+  failure_threshold = 1
```
acsauk, line 27: Was there a reason for making this an instant fail? Could we maybe increase the threshold and decrease the request interval to account for blips (unless we care about blips (such a technical term 😆)?)

andrewpearce-digital: I don't know what the pattern for dependency failures would look like, so starting here and we can adjust as we learn more.

PR #923, opened 2023-12-13
MLPAB-1700 - Create a health check SNS topic with encryption

andrewpearce-digital:
# Purpose

Share an SNS topic with encryption

Fixes MLPAB-1700

## Approach

- create an account level SNS topic for health check alarms


acsauk (approved):
:shipit:

PR #926, opened 2023-12-14
MLPAB-1637 Add the other data to the lpa store submission

hawx:
Also see https://github.com/ministryofjustice/opg-data-lpa-store/pull/72

### DIFF
--- a/internal/lpastore/client_test.go
+++ b/internal/lpastore/client_test.go
@@ -51,6 +51,7 @@ func TestClientSendLpa(t *testing.T) {
 
 	donor := &actor.DonorProvidedDetails{
 		LpaUID: "M-0000-1111-2222",
+		Type:   actor.LpaTypeHealthWelfare,
 		Donor: actor.Donor{
 			FirstNames:  "John Johnson",
 			LastName:    "Smith",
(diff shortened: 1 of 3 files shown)


internal/lpastore/client_test.go
```diff
@@ -241,9 +279,11 @@ func TestClientServiceContract(t *testing.T) {
 				"X-Jwt-Authorization": dsl.Regex("Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJvcGcucG9hcy5tYWtlcmVnaXN0ZXIiLCJzdWIiOiJ0b2RvIiwiaWF0Ijo5NDY3NzEyMDB9.teh381oIhucqUD3EhBTaaBTLFI1O2FOWGe-44Ftk0LY", "Bearer .+"),
 			},
 			Body: dsl.Like(map[string]any{
```
acsauk, line 281: Should this include `CertificateProvider`?

acsauk: And is it worth including `PeopleToNotify` or are we going for minimum required data?

acsauk (commented):
From looking at the API spec it looks like we need `certificateProvider` as part of the `initialLpa` unless I'm misreading the validation blocks

internal/lpastore/client_test.go
hawx: good points

PR #927, opened 2023-12-14
MLPAB-1704 - create kms key alias in global region

andrewpearce-digital:
# Purpose

Encrypt SNS topic in all regions it is used

Fixes MLPAB-1704

## Approach

- create kms key alias in global region


acsauk (approved):
:shipit:

PR #929, opened 2023-12-15
MLPAB-1404: Prevent donor signing when attorneys under 18

acsauk:
# Purpose

As noted in the ticket, we're leaving CPs for now as this data isn't provided by the donor.

Fixes MLPAB-1404


internal/date/date.go
```diff
@@ -22,7 +22,7 @@ type Date struct {
 	month string
 	day   string
 
-	t   time.Time
+	T   time.Time
```
acsauk, line 25: Exporting this so when we check object hashes at the start of a `Put` there is something to compare against.

hawx: I'd prefer to do https://pkg.go.dev/github.com/mitchellh/hashstructure#Hashable, which could be as simple as 

```
func (d Date) Hash() (uint64, error) { return uint64(d.t.Unix()), nil }
```

web/template/you_cannot_sign_your_lpa_yet.gohtml
```diff
@@ -0,0 +1,54 @@
...
+            {{ template "warning-banner" (warning .App "actorsMustBeOver18WarningBanner") }}
+
+            <h1 class="govuk-heading-xl">{{ tr .App "youCannotSignYourLpaYet" }}</h1>
+
+            {{ $actorsData := .Donor.Under18ActorDetails }}
+            {{ $from := printf "%s%s" $.App.Page "#f-date-of-birth" }}
+
+            {{ if eq 1 (len $actorsData) }}
+                {{ trHtml .App "singularActorYouCannotSignYourLpaYetContent" }}
+                {{ $from = .App.Paths.TaskList.Format .App.LpaID }}
```
hawx, line 17: why change this in the single case?

hawx: and if you don't need to change this you could use `{"one": ..., "other": ...}` for the content instead of an `if` here

web/template/you_cannot_sign_your_lpa_yet.gohtml
```diff
@@ -0,0 +1,54 @@
...
+                            </a>
+                        </dd>
+                    </div>
+                {{ end }}
+            </dl>
+
+            {{ if eq 1 (len $actorsData) }}
+                {{ trHtml .App "singularActorIfDateYouEnteredIsCorrectYouMustWaitContent" }}
+            {{ else }}
+                {{ trHtml .App "multipleActorsIfDateYouEnteredIsCorrectYouMustWaitContent" }}
```
hawx, line 48: could use `trFormatCount` here, and remove the HTML from the content since it is just a `<p>`

hawx (commented):
there is also a massive assumption on the ticket that when we check ID we make sure someone is 18, which we do not do

web/template/you_cannot_sign_your_lpa_yet.gohtml
acsauk: I was trying my best not to have to write a handler here so I hacked in some redirect logic for cases where the donor changes DOBs or removes actors and lands back on this page with 0 actors. Now I'm typing this out though it doesn't account for changing the DOB to something that is still under 18 so it probably makes sense to put a simple handler in to check this for redirects.

And cool, I'd forgotten about the pluralize feature.

hawx (approved):

PR #930, opened 2023-12-15
MLPAB-1712: Work out why CPs on demo are being redirected to the dashboard

acsauk:
# Purpose

I've not been able to test this on a PR env due to the redirect timing out (assuming there were some env vars I failed to update) but testing this again on demo and on starting the CP journey and leaving the sub value empty in the mock is now consistently landing on the reference number page. 

My hunch was the randomness was potentially not as random as we needed so I've upgraded to subs to uuids as some protection against this.

Fixes MLPAB-1712 (maybe?)

### DIFF
--- a/internal/page/fixtures/attorney.go
+++ b/internal/page/fixtures/attorney.go
@@ -8,6 +8,7 @@ import (
 	"slices"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/ministryofjustice/opg-go-common/template"
 	"github.com/ministryofjustice/opg-modernising-lpa/internal/actor"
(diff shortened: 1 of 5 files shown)


internal/page/fixtures/attorney.go
```diff
@@ -64,7 +65,7 @@ func Attorney(
 		)
 
 		if attorneySub == "" {
-			attorneySub = random.String(16)
```
hawx, line 67: I really doubt that 62^16 isn't enough, so would be tempted to say put these back so we can see the actual change here

acsauk:
As demo has started to magically work we're closing this ticket. We can revisit if problems occur again.

PR #933, opened 2023-12-19
MLPAB-1637 Rename enum string values to match what we are sending to lpastore

hawx:


acsauk (approved):
:shipit: 

Tidy. I like the new approach for translation keys that vary by type (or something).

PR #935, opened 2023-12-19
MLPAB-1650, MLPAB-1651: Update and always show DOB warning, update name warning

acsauk:
# Purpose

Fixes MLPAB-1650, MLPAB-1651.


internal/page/donor/choose_attorneys.go
```diff
@@ -59,7 +60,9 @@ func ChooseAttorneys(tmpl template.Template, donorStore DonorStore, uuidString f
 				data.DobWarning = dobWarning
 			}
 
-			if data.Errors.Any() || data.Form.IgnoreNameWarning != nameWarning.String() {
+			if data.Errors.Any() ||
+				data.Form.IgnoreNameWarning != nameWarning.String() &&
+					attorney.FullName() != fmt.Sprintf("%s %s", data.Form.FirstNames, data.Form.LastName) {
```
hawx, line 65: I feel like this would make more sense as `attorney.FirstNames != data.Form.FirstNames && attorney.LastName != data.Form.LastName` (and, not that I think people would ever do this, at the moment going from [A B] [C] to [A] [B C] would look like no change)

hawx: also, the logic should be `nameHasntChanged && (anyErrors || warningNotIgnored)` otherwise if you:
1. make an attorney have the same name as the donor
2. accept the warning
3. go back to that attorney and make the dob invalid
4. you get the same name warning again

acsauk:
Closing in favour of #948 (due to tainted pager duty resource)

PR #940, opened 2023-12-22
MLPAB-1638 Send certificate provider info to lpa store

hawx:


internal/lpastore/client.go
```diff
@@ -214,6 +214,68 @@ func (c *Client) SendLpa(ctx context.Context, donor *actor.DonorProvidedDetails)
...
+}
+
+type updateRequest struct {
+	Type    string                `json:"type"`
+	Changes []updateRequestChange `json:"changes"`
+}
+
+type updateRequestChange struct {
+	Key string `json:"key"`
+	Old any    `json:"old,omitempty"`
```
acsauk, line 228: We don't appear to be using this (yet) but the [API spec](https://github.com/ministryofjustice/opg-data-lpa-store/blob/main/docs/openapi/openapi.yaml#L454) states it is a required field. Is this in anticipation of a future requirement?

acsauk (approved):
:shipit:

PR #961, opened 2024-01-15
MLPAB -1378: Add personal details warnings for subsequent lpas

acsauk:
# Purpose

Fixes MLPAB-1378

### DIFF
--- a/docker/localstack/Dockerfile
+++ b/docker/localstack/Dockerfile
@@ -1,6 +1,22 @@
+FROM golang:1.21.5-alpine as build
+
+WORKDIR /app
+
+COPY --link go.mod go.sum ./
+RUN go mod download
+
+COPY --link cmd/event-received ./cmd/event-received
+COPY --link internal ./internal
+
+RUN GOOS=linux GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -tags lambda.norpc -o event-received ./cmd/event-received
+
 FROM localstack/localstack:3.0.2 as localstack
 
-COPY localstack-init.sh /etc/localstack/init/ready.d/localstack-init.sh
-COPY dynamodb-lpa-gsi-schema.json ./dynamodb-lpa-gsi-schema.json
+COPY --from=build /app/event-received ./event-received
+RUN zip event-received.zip event-received
+RUN mv event-received.zip /etc/event-received.zip
+
+COPY ./docker/localstack/localstack-init.sh /etc/localstack/init/ready.d/localstack-init.sh
+COPY ./docker/localstack/dynamodb-lpa-gsi-schema.json ./dynamodb-lpa-gsi-schema.json
 
 RUN chmod 544 /etc/localstack/init/ready.d/localstack-init.sh
--- a/internal/page/dashboard.go
+++ b/internal/page/dashboard.go
@@ -32,15 +32,6 @@ type dashboardData struct {
 
 func Dashboard(tmpl template.Template, donorStore DonorStore, dashboardStore DashboardStore) Handler {
 	return func(appData AppData, w http.ResponseWriter, r *http.Request) error {
-		if r.Method == http.MethodPost {
-			lpa, err := donorStore.Create(r.Context())
-			if err != nil {
-				return err
-			}
-
-			return Paths.YourDetails.Redirect(w, r, appData, lpa)
-		}
-
 		donorLpas, attorneyLpas, certificateProviderLpas, err := dashboardStore.GetAll(r.Context())
 		if err != nil {
 			return err
@@ -55,6 +46,20 @@ func Dashboard(tmpl template.Template, donorStore DonorStore, dashboardStore Das
 		}
 		if len(attorneyLpas) > 0 {
 			tabCount++
+		}
+
+		if r.Method == http.MethodPost {
+			lpa, err := donorStore.Create(r.Context())
+			if err != nil {
+				return err
+			}
+
+			path := Paths.YourDetails
+			if len(donorLpas) > 0 {
+				path = Paths.MakeANewLPA
+			}
+
+			return path.Redirect(w, r, appData, lpa)
 		}
 
 		data := &dashboardData{
--- a/internal/page/donor/make_a_new_lpa.go
+++ b/internal/page/donor/make_a_new_lpa.go
@@ -1,0 +1,43 @@
+package donor
+
+import (
+	"net/http"
+
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/actor"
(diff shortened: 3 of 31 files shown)


docker/localstack/Dockerfile
```diff
@@ -1,6 +1,22 @@
...
+COPY --link internal ./internal
+
+RUN GOOS=linux GOOS=${TARGETOS} GOARCH=${TARGETARCH} CGO_ENABLED=0 go build -tags lambda.norpc -o event-received ./cmd/event-received
+
 FROM localstack/localstack:3.0.2 as localstack
 
-COPY localstack-init.sh /etc/localstack/init/ready.d/localstack-init.sh
-COPY dynamodb-lpa-gsi-schema.json ./dynamodb-lpa-gsi-schema.json
+COPY --from=build /app/event-received ./event-received
+RUN zip event-received.zip event-received
```
hawx, line 16: you could move this step to the other stage to make this image smaller

internal/page/dashboard.go
```diff
@@ -56,6 +47,20 @@ func Dashboard(tmpl template.Template, donorStore DonorStore, dashboardStore Das
...
 		}
 
+		if r.Method == http.MethodPost {
+			lpa, err := donorStore.Create(r.Context())
+			if err != nil {
+				return err
+			}
+
+			path := Paths.YourDetails
+			if len(donorLpas) > 0 {
```
hawx, line 57: Alternatively, could add a hidden field to the template if donorLpas>0 then check that on POST. That would save doing the dashboardStore.GetAll query when making a new one.

internal/page/donor/make_a_new_lpa.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/ministryofjustice/opg-go-common/template"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/actor"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/date"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/page"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/place"
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/validation"
+)
+
+type MakeANewLPAData struct {
```
hawx, line 14: lowercase m here to be consistent on keeping these private

internal/page/donor/make_a_new_lpa.go
```diff
@@ -0,0 +1,66 @@
...
+		data := MakeANewLPAData{
+			App:         appData,
+			FullName:    previouslyProvidedDetails.Donor.FullName(),
+			DateOfBirth: previouslyProvidedDetails.Donor.DateOfBirth,
+			Address:     previouslyProvidedDetails.Donor.Address,
+		}
+
+		if data.FullName != donor.Donor.FullName() && donor.Donor.FirstNames != "" {
+			data.FullName = donor.Donor.FullName()
+		} else {
```
hawx, line 38: I may be reading this wrong, but I don't think the `else`s do anything and you don't need the `donorStore.Put`

hawx: which would also mean you could move the `if POST { }` to the top

internal/page/donor/make_a_new_lpa_test.go
```diff
@@ -0,0 +1 @@
+package donor
```
hawx, line 1: :wave:

internal/page/donor/your_address.go
```diff
@@ -42,6 +55,11 @@ func YourAddress(logger Logger, tmpl template.Template, addressClient AddressCli
 						return err
 					}
 
+					if makingAnotherLPA {
+						r.Form.Del("from")
```
hawx, line 59: I'd just use a different param for making another, that way `from` can keep working the same way

internal/page/donor/your_name.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		if r.Method == http.MethodPost {
+			loginSession, err := sesh.Login(sessionStore, r)
+			if err != nil {
+				return err
+			}
+			if loginSession.Email == "" {
+				return fmt.Errorf("no email in login session")
+			}
```
hawx, line 51: could this be left out here, since you'll pick it up when you Continue and go through /your-details?

internal/page/donor/your_name.go
```diff
@@ -0,0 +1,131 @@
...
+			)
+
+			if data.Errors.Any() ||
+				data.Form.IgnoreNameWarning != nameWarning.String() &&
+					donor.Donor.FullName() != fmt.Sprintf("%s %s", data.Form.FirstNames, data.Form.LastName) {
+				data.NameWarning = nameWarning
+			}
+
+			if !data.Errors.Any() && data.NameWarning == nil {
+				changesMade := donor.Donor.FirstNames != data.Form.FirstNames || donor.Donor.LastName != data.Form.LastName
```
hawx, line 70: just to check, are you purposefully not checking other names here?

hawx: also, this doesn't work if you do:
- create registered LPA in fixtures
- go to create a new LPA
- go to change name
- don't edit the name, just continue
- see the "we have updated your details" page (but nothing has changed...)

internal/page/dashboard.go
```diff
@@ -56,6 +47,20 @@ func Dashboard(tmpl template.Template, donorStore DonorStore, dashboardStore Das
 			tabCount++
 		}
 
+		if r.Method == http.MethodPost {
+			lpa, err := donorStore.Create(r.Context())
```
hawx, line 51: Would it be simpler to add the details to the new LPA at this point vs. doing `.Latest` in all the handlers? We could even have a `donorStore.CreateFromLatest` or something, or just make it check in `.Create` whether one exists?

internal/page/donor/your_name.go
hawx: and should there be a similar check on the date of birth and address pages?

docker/localstack/Dockerfile
acsauk: I was being lazy as the base and alpine Go images don't have zip installed but agree makes sense to not double the size in the image.

internal/page/dashboard.go
acsauk: Much, I'll move to here.

internal/page/donor/your_name.go
acsauk: Totally had the intentions of adding to the other handlers - should be working now

internal/page/donor/make_a_new_lpa_test.go
acsauk: oopsie - I've dropped the hnadler now anyway as we can get by with using `Guidance`

hawx (approved):

PR #963, opened 2024-01-16
MLPAB-1491 Use 12-hour clock for times

hawx:


acsauk (approved):
Sophie will be happy :shipit:

PR #964, opened 2024-01-16
MLPAB-1484 Escape data passed to templated translation messages

hawx:
https://github.com/nicksnyder/go-i18n/blob/main/v2/internal/template.go#L7 is the line that we'd need to change, but given we don't actually use much of the package I thought it might be easier writing this. It also opens up a bit more flexibility, e.g. if we wanted to simplify Welsh plural forms in the future.


acsauk (approved):
Nice job, hopefully we can look at how often (if ever) our translations fall into the 4 variations of pluralisms for Welsh and slim that down if it turns out they're the same everywhere.

:shipit:

PR #965, opened 2024-01-17
MLPAB-1757: Extract field name and options to YesNoForm

acsauk:
# Purpose

And a further follow on from #966, this time applying to the `YesNo` form. I've moved the option values into the form as this reduced some manual injections into data objects that relied on them.

Fixes MLPAB-1757


internal/form/yes_no.go
```diff
@@ -4,6 +4,7 @@ package form
 type YesNo uint8
```
hawx, line 4: with the explicit unknown it would be better to remove `-empty` and add `-trimprefix`

internal/form/yes_no_form.go
```diff
@@ -11,22 +11,26 @@ type YesNoForm struct {
...
 func ReadYesNoForm(r *http.Request, errorLabel string) *YesNoForm {
-	yesNo, err := ParseYesNo(PostFormString(r, "yes-no"))
+	form := NewYesNoForm(YesNoUnknown)
 
-	return &YesNoForm{
-		YesNo:      yesNo,
-		Error:      err,
-		ErrorLabel: errorLabel,
-	}
+	yesNo, err := ParseYesNo(PostFormString(r, form.FieldName))
```
hawx, line 21: if you moved this line up you could do `form := NewYesNoForm(yesNo)`

acsauk: I could, but then we'd have to reference the field name directly rather than via the Form object. While that's not too terrible given where we are referencing it, it does break the encapsulation. What do you think?

hawx: couldn't you use `FieldNames.YesNo`?

hawx (approved):
but this is also fine

PR #966, opened 2024-01-17
MLPAB-1757: Extract address field names

acsauk:
# Purpose

Following on from keeping shared field names centrally defined in #894 this tackles the address form. I went through a couple of variations of how they are made available to the data objects that use them and injecting into the form object itself felt like the least burdensome as far as maintenance and testing are concerned.

Fixes MLPAB-1757


internal/form/names.go
```diff
@@ -1,9 +1,31 @@
 package form
 
-var FieldNames = Names{
-	LanguagePreference: "language-preference",
+var FieldNames = SharedFieldNames{
+	LanguagePreference: LanguagePreferenceFieldNames{
```
hawx, line 4: I think this would be ok as it was, since it doesn't have multiple fields (and I don't like the repetition it has introduced)

internal/form/names.go
```diff
@@ -1,9 +1,31 @@
...
+		LanguagePreference: "language-preference",
+	},
+	Address: AddressFieldNames{
+		Line1:      "address-line-1",
```
hawx, line 8: probably should add `action` here too, since it is a shared field

hawx (approved):

PR #969, opened 2024-01-18
MLPAB-1733 - Make the public access toggle include redirects and mock onelogin

andrewpearce-digital:
# Purpose

Manage all ingress routes when toggling public access to the UR environment

Fixes MLPAB-1733

## Approach

- update targets for terraform apply on the UR environment public access workflow
- remove private ingress rules when public access enabled and vice versa

Without the toggling of rule sets, we exceed the number of rules allowed


acsauk (approved):
:shipit:

PR #972, opened 2024-01-19
MLPAB-1740 Add page to enter an organisation or company name

hawx:


internal/actor/group.go
```diff
@@ -0,0 +1,26 @@
+package actor
+
+import "time"
+
+// A Group contains users associated with a set of permissions that work on the
+// same set of LPAs.
+type Group struct {
+	PK, SK string
```
mattmachell, line 8: The only reticence I have with calling it "group" is that isn't the term that's been used by business/product/user folks, so might lead to confusion later on down the line when discussing this.

andrewpearce-digital (approved):

andrewpearce-digital (approved):

PR #977, opened 2024-01-23
MLPAB-1738 Add page to invite a member to an organisation

hawx:


internal/notify/email.go
```diff
@@ -111,3 +111,16 @@ func (e CertificateProviderProvideCertificatePromptEmail) WithShareCode(shareCod
...
 }
+
+type MemberInviteEmail struct {
+	OrganisationName string
+	InviteCode       string
+}
+
+func (e MemberInviteEmail) emailID(isProduction bool) string {
+	if isProduction {
+		return "-"
```
hawx, line 122: we don't have emails yet so left it like this

acsauk (approved):
:shipit:

PR #979, opened 2024-01-24
Remove LegacyString method from LpaType

hawx:
For when https://github.com/ministryofjustice/opg-sirius/pull/9707 is merged


townxelliot:
https://github.com/ministryofjustice/opg-sirius/pull/9707 is merged, so this should be OK to merge.

acsauk (commented):
Do we want to update any related tests still using the legacy strings in LPA type or are there still some event receivers / endpoints that require the abbreviated types? I'm thinking specifically `handleUidRequested()` in `event-received` and client tests for the UID client (and if so then may as well update the `emit-uid-requested` make command while we're at it).

acsauk (approved):
:shipit:

hawx:
https://github.com/ministryofjustice/opg-data-lpa-uid/pull/129 is when the UID service was updated

PR #1012, opened 2024-01-31
MLPAB-1806 Add confirmation page for when a member is invited

hawx:


internal/page/fixtures/supporter.go
```diff
@@ -1,25 +1,40 @@
...
 
 		if err := sesh.SetLoginSession(sessionStore, r, w, &sesh.LoginSession{Sub: supporterSub, Email: testEmail}); err != nil {
 			return err
 		}
 
+		if organisation == "1" {
+			if err := organisationStore.Create(ctx, random.String(12)); err != nil {
+				return err
+			}
+		}
```
acsauk, line 36: Are we expecting scenarios where supporters won't be part of an organisation? Or is this more to allow for scaffolding pre-org-creation?

acsauk (approved):
:shipit:

internal/page/fixtures/supporter.go
hawx: the other test we have is creating the organisation so doesn't use this, but you are right we may end up flipping this as we add other tests

PR #1016, opened 2024-02-01
MLPAB-1781 Show organisation name in header

hawx:


internal/app/app.go
```diff
@@ -267,7 +266,7 @@ func makeHandle(mux *http.ServeMux, errorHandler page.ErrorHandler, store sesh.S
 					return
 				}
 
-				appData.SessionID = base64.StdEncoding.EncodeToString([]byte(loginSession.Sub))
+				appData.SessionID = loginSession.SessionID()
```
hawx, line 269: I'm not sure why we haven't done this sooner, it has ended up in so many places

acsauk: Haha I've just been further sprinkling the codebase with the long form of this in my PR - good spot

acsauk (approved):
:shipit:

PR #1017, opened 2024-02-01
MLPAB-1743: Create a new LPA as a supporter

acsauk:
# Purpose

The bulk of change here is getting the required data in session to make the jump from supporter to donor handlers. There will need to be a follow-up ticket to fix any issues encountered in the donor flow while acting as a supporter.

Fixes MLPAB-1743


internal/app/organisation_store.go
```diff
@@ -93,6 +111,37 @@ func (s *organisationStore) CreateMemberInvite(ctx context.Context, organisation
 	return nil
 }
 
+func (s *organisationStore) CreateLPA(ctx context.Context, organisationID string) (*actor.DonorProvidedDetails, error) {
```
acsauk, line 114: I'm leaving this here for now. We could probably modify `donorStore.Create()` but I didn't want to have to think through the `LpaLink` implications until we need it for the dashboard.

web/assets/scss/main.scss
```diff
@@ -98,6 +98,7 @@ body:not(.js-enabled) .govuk-back-link {
     @include govuk-media-query($from: tablet) {
         @include govuk-responsive-padding(4, "bottom");
     }
+    color: govuk-colour("black");
```
acsauk, line 101: Following design review

internal/page/supporter/login_callback.go
```diff
@@ -35,21 +35,31 @@ func LoginCallback(oneLoginClient LoginCallbackOneLoginClient, sessionStore sesh
...
-		if err := sesh.SetLoginSession(sessionStore, r, w, session); err != nil {
-			return err
-		}
-
 		ctx := page.ContextWithSessionData(r.Context(), &page.SessionData{SessionID: session.SessionID()})
 
-		_, err = organisationStore.Get(ctx)
+		organisation, err := organisationStore.Get(ctx)
 		if err == nil {
+			session.OrganisationID = organisation.ID
```
hawx, line 46: We could extend session with `OrganisationName` too, then we can let that flow through to AppData so the header contains it. Having it not be "live" is an ok tradeoff for now?

internal/page/supporter/register.go
```diff
@@ -45,23 +46,23 @@ type ErrorHandler func(http.ResponseWriter, *http.Request, error)
...
+	supporterTmpls template.Templates,
 	oneLoginClient OneLoginClient,
 	sessionStore SessionStore,
 	organisationStore OrganisationStore,
 	notFoundHandler page.Handler,
 	errorHandler page.ErrorHandler,
 	notifyClient NotifyClient,
 ) {
-	paths := page.Paths.Supporter
+	supporterPaths := page.Paths.Supporter
```
hawx, line 57: you should be able to rename these back

hawx (dismissed):

internal/page/supporter/login_callback.go
acsauk: Yeah thats probably a good idea, I'll add in.

hawx (approved):

PR #1020, opened 2024-02-01
MLPAB-1839 - reinstate account level replicatoin role

andrewpearce-digital:
# Purpose

This role was moved to environment level because we exceeded the limit of policies attached to an IAM role (10). That move has disrupted to configuration of Sirius, and made it quite difficult to tighten up.

I've increased the limit to 20, which should be more than enough.

Fixes MLPAB-1839

## Approach

- Create account level replication role

## Learning

- https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html


acsauk (approved):
:shipit:

PR #1025, opened 2024-02-06
MLPAB-1542: Emit paper-forms-requested event for CPs acting on paper

acsauk:
# Purpose

Fixes MLPAB-1542


internal/page/share_code.go
```diff
@@ -49,6 +50,13 @@ func (s *ShareCodeSender) SendCertificateProviderInvite(ctx context.Context, app
 }
 
 func (s *ShareCodeSender) SendCertificateProviderPrompt(ctx context.Context, appData AppData, donor *actor.DonorProvidedDetails) error {
+	if donor.CertificateProvider.CarryOutBy.IsPaper() {
+		return s.eventClient.SendPaperFormRequested(ctx, event.PaperFormRequested{
+			UID:       donor.LpaUID,
+			ActorType: actor.TypeCertificateProvider.String(),
+		})
+	}
```
acsauk, line 58: While this technically isn't a share code being sent, it is inviting the CP to start their process and feels like a nice place to contain this logic. If we add to this further this could be renamed to something like `ActorInviter`?

hawx: yeah, it might be nice one day to pull all of this out to an `invite` package and stop using the words "share code"

hawx (approved):

PR #1028, opened 2024-02-08
MLPAB-1866 - include all targets for public access toggling

andrewpearce-digital:
# Purpose

Toggling public access was failing because the rule limit was exceeded. This was because it wasn't removing private ingress rules while adding public ingress rules, due to missing targets in the targeted terraform apply

Fixes MLPAB-1866

## Approach

- include all targets required for toggling access


acsauk (approved):
:shipit:

PR #1031, opened 2024-02-08
MLPAB-1833: Add name and permissions to invite member

acsauk:
# Purpose

Fixes MLPAB-1833


internal/page/supporter/invite_member.go
```diff
@@ -50,21 +60,40 @@ func InviteMember(tmpl template.Template, organisationStore OrganisationStore, n
...
+
+	errors.String("last-name", "lastName", f.LastName,
+		validation.Empty(),
+		validation.StringTooLong(61))
+
 	errors.String("email", "email", f.Email,
 		validation.Empty(),
 		validation.Email())
 
+	errors.Options("permission", "makeThisPersonAnAdmin", []string{f.Permission.String()}, validation.Select(actor.None.String(), actor.Admin.String()))
```
acsauk, line 96: This is a bit of a hack for now - we've spoken about having different granular permissions in the future but it didn't make sense to make `Permission` -> `Permissions` with only a single option.

internal/actor/permission.go
```diff
@@ -0,0 +1,9 @@
+package actor
+
+//go:generate enumerator -type Permission -linecomment
+type Permission uint8
+
+const (
+	None  Permission = iota //none
+	Admin                   // admin
```
hawx, line 8: PermissionNone and PermissionAdmin to make it more clear? (and then `-trimprefix` on the enumerator call)

hawx: I won't make the permissions/role comment. This rename can also be picked up when we change these later

hawx (approved):

PR #1034, opened 2024-02-09
MLPAB-1742 Add list of LPAs to supporter dashboard

hawx:


internal/app/organisation_store.go
```diff
@@ -131,6 +133,32 @@ func (s *organisationStore) CreateLPA(ctx context.Context, organisationID string
 	return donor, err
 }
 
+func (s *organisationStore) AllLPAs(ctx context.Context) ([]actor.DonorProvidedDetails, error) {
```
hawx, line 136: not the most efficient query, but will all be replaced as soon as we touch a story with filtering/searching

acsauk (approved):
:shipit:

PR #1038, opened 2024-02-12
MLPAB-1894 Add UID to actors

hawx:
I think having its own package makes sense as I want to avoid throwing yet more stuff in the `actor` bag, and at some point we should introduce a type for LPA UID to tighten that up and this avoids naming confusion


acsauk (approved):
:shipit:

PR #1040, opened 2024-02-13
Copy in the only thing we are using golang/x/exp for

hawx:
I'm bored of seeing extra renovate PRs for golang.org/x/exp, like https://github.com/ministryofjustice/opg-modernising-lpa/pull/1037, when we only have the package for a single little helper that may one day be in the standard library


acsauk (approved):
:shipit:

PR #1042, opened 2024-02-14
Use log/slog for logging

hawx:
`log/slog` has been in the standard library for a bit now, and changing to it will give us more flexibility in what we log vs. the `opg-go-common/logging` that was just pulled out of our first Go project


andrewpearce-digital (approved):

acsauk (approved):
:shipit:

PR #1046, opened 2024-02-14
MLPAB-1746: Create and show Member when accepting Organisation invite

acsauk:
# Purpose

Fixes MLPAB-1746


.pre-commit-config.yaml
```diff
@@ -53,8 +53,8 @@ repos:
     rev: 37.150.1
     hooks:
       - id: renovate-config-validator
-  - repo: https://github.com/Yelp/detect-secrets
-    rev: v1.4.0
-    hooks:
-    -   id: detect-secrets
-        args: ['--baseline', '.secrets.baseline']
+#  - repo: https://github.com/Yelp/detect-secrets
```
acsauk, line 56: Didn't have the energy to figure out whats broken with my latest package updates with this - will tackle in a separate PR

internal/app/dashboard_store.go
```diff
@@ -23,7 +23,7 @@ type lpaLink struct {
 	DonorKey string
 	// ActorType is the type for the current user
 	ActorType actor.Type
-	// UpdatedAt is set to allow this data to be queried from ActorUpdatedAtIndex
+	// UpdatedAt is set to allow this data to be queried from SkUpdatedAtIndex
```
acsauk, line 26: One downside of relying on this index but not the end of the world.

internal/actor/organisation.go
```diff
@@ -28,13 +28,26 @@ type Member struct {
...
+	return fmt.Sprintf("%s %s", i.FirstNames, i.LastName)
 }
 
 // A MemberInvite is created to allow a new Member to join an Organisation
 type MemberInvite struct {
 	PK, SK string
 	// CreatedAt is when the MemberInvite was created
 	CreatedAt time.Time
+	// UpdatedAt is when the MemberInvite was last updated
+	UpdatedAt time.Time
```
acsauk, line 50: One downside of relying on`SkUpdatedAtIndex` but not the end of the world.

cmd/mock-onelogin/main.go
```diff
@@ -173,6 +180,8 @@ func authorize() http.HandlerFunc {
 			sub:      r.FormValue("sub"),
 		}
 
+		emailOverride = r.FormValue("email")
```
hawx, line 183: this could live in `sessionData` to avoid having the global state

internal/actor/organisation.go
```diff
@@ -28,13 +28,26 @@ type Member struct {
...
+	LastLoggedInAt time.Time
+	Email          string
+	FirstNames     string
+	LastName       string
+	// Permission is the type of permissions assigned to the member to set available actions in an Organisation
+	Permission Permission
+}
+
+func (i Member) FullName() string {
+	return fmt.Sprintf("%s %s", i.FirstNames, i.LastName)
```
hawx, line 41: can you add a simple test for this

internal/actor/organisation.go
hawx: it's fine, we should be consistent with these, even if we aren't changing an entity

internal/app/app.go
```diff
@@ -44,7 +44,7 @@ type DynamoClient interface {
 	OneByPartialSk(ctx context.Context, pk, partialSk string, v interface{}) error
 	AllByPartialSk(ctx context.Context, pk, partialSk string, v interface{}) error
 	LatestForActor(ctx context.Context, sk string, v interface{}) error
-	AllForActor(ctx context.Context, sk string, v interface{}) error
+	AllBySK(ctx context.Context, sk string, v interface{}) error
```
hawx, line 47: do you prefer the naming `AllBySK` or `AllBySk`? would be good to be consistent either way, and if it is changing the others to uppercase leave it for a separate PR

internal/page/supporter/enter_reference_number.go
```diff
@@ -0,0 +1,88 @@
...
+				if err := organisationStore.CreateMember(r.Context(), invite); err != nil {
+					return err
+				}
+
+				loginSession, err := sesh.Login(sessionStore, r)
+				if err != nil {
+					return page.Paths.Supporter.Start.Redirect(w, r, appData)
+				}
+
+				loginSession.OrganisationID = invite.OrganisationID
```
hawx, line 51: should also `loginSession.OrganisationName = invite.OrganisationName`, or get the organisation to use the current name

internal/app/app.go
acsauk: We spoke a while back about capitalising so let's go with that. I'll pick up next.

cmd/mock-onelogin/main.go
acsauk: I thought this would be trickier to get back out that it turned out being - good call.

internal/app/organisation_store.go
```diff
@@ -44,8 +49,9 @@ func (s *organisationStore) Create(ctx context.Context, name string) (*actor.Org
 	}
 
 	member := &actor.Member{
-		PK:        memberKey(data.SessionID),
-		SK:        organisationKey(organisationID),
+		PK:        organisationKey(organisationID),
+		SK:        memberKey(data.SessionID),
+		Email:     data.Email,
```
hawx, line 54: should we be updating this if it has changed when they login?

web/template/supporter/manage_team_members.gohtml
```diff
@@ -75,6 +75,54 @@
...
                         </table>
                     {{ end }}
+
+                    {{ if .Members }}
+                        <h2 class="govuk-heading-m">{{ tr .App "teamMembers" }}</h2>
+
+                        <table class="govuk-table">
+                            <thead class="govuk-table__head">
+                            <tr class="govuk-summary-list__row">
+                                <th class="govuk-table__header govuk-!-font-weight-bold">
```
hawx, line 85: these are bold by default

web/template/supporter/enter_reference_number.gohtml
```diff
@@ -0,0 +1,21 @@
...
+{{ define "pageTitle" }}{{ tr .App "enterYourSupporterReferenceNumber" }}{{ end }}
+
+{{ define "main" }}
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <h1 class="govuk-heading-xl">{{ tr .App "enterYourSupporterReferenceNumber" }}</h1>
+
+            <form novalidate method="post">
+                <div class="govuk-form-group">
+                    <fieldset class="govuk-fieldset">
```
hawx, line 12: you don't need the `.govuk-form-group` or `fieldset` wrappers

hawx (commented):
it would be good to add coverage where codecov says it is missing

internal/app/organisation_store.go
acsauk: Agreed we are updating on login for supporters but will discuss with product if we care about GOL email changes.

hawx (approved):

PR #1054, opened 2024-02-16
MLPAB-1913 Fix translations where ActorType was previously dynamic

hawx:


acsauk (commented):
Unlikely to happen but is it worth adding a test for an unexpected actor type using `ChangeMobileNumber()`?

acsauk (approved):
:shipit:

PR #1058, opened 2024-02-19
MLPAB-1679 Create .app-tag with no max-width

hawx:
I think using `.app-tag` instead of adding a modifier makes more sense, otherwise we will end up with really long classes


acsauk (approved):
:shipit:

PR #1064, opened 2024-02-20
MLPAB-1906- create missing workspaces so that the destroy job never fails

andrewpearce-digital:
# Purpose

Prevent the PR environment destroy job failing when a workspace doesn't exist

## Approach

- Create a workspace if it doesn't exist

This lets the rest of the job pass very quickly


hawx (approved):

acsauk (approved):
:shipit:

PR #1070, opened 2024-02-22
MLPAB-1785 Add opensearch and pagination to supporter dashboard

hawx:


terraform/environment/opensearch.tf
```diff
@@ -0,0 +1,114 @@
...
+
+  filter {
+    name   = "tag:Name"
+    values = ["application*"]
+  }
+  provider = aws.eu_west_1
+}
+
+resource "aws_opensearchserverless_security_policy" "lpas_collection_encryption_policy" {
+  name        = "policy-${local.environment_name}"
```
andrewpearce-digital, line 35: these resources are missing providers and will fail to create on the pipeline (as there's no default provider, the user in identity will try to create these in us-east-1 but it doesn't have permission to do that)

terraform/environment/regions.tf
```diff
@@ -12,6 +12,11 @@ module "allow_list" {
   source = "git@github.com:ministryofjustice/opg-terraform-aws-moj-ip-allow-list.git?ref=v2.3.0"
 }
 
+data "aws_opensearchserverless_collection" "lpas_collection" {
```
andrewpearce-digital, line 15: we shouldn't use the data source here because it will look for it at the refresh stage before it's created and will fail. instead use the attribute of the resource `aws_opensearchserverless_collection.lpas_collection.collection_endpoint`

terraform/environment/region/variables.tf
```diff
@@ -152,3 +152,8 @@ variable "fault_injection_experiments_enabled" {
   type        = bool
   description = "Enable fault injection"
 }
+
+variable "search_endpoint" {
+  type        = string
+  description = "URL of the OpenSearch Service endpoint to use"
```
andrewpearce-digital, line 158: Because we know that there's no service in London, perhaps we can make this nullable for now? https://developer.hashicorp.com/terraform/language/values/variables#disallowing-null-input-values

internal/search/pagination_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	for current, pages := range testCases {
+		t.Run(fmt.Sprintf("Page%d", current), func(t *testing.T) {
+			pagination := newPagination(250, current, 25)
+
+			assert.Equal(t, pages, pagination.Pages())
+		})
+	}
+}
```
acsauk, line 131: This has saved some confused questions on adding -1 to the pages slice, thanks 🙏

acsauk (approved):
Nice job :shipit:

cmd/event-received/cloud_watch_event_handler.go
```diff
@@ -32,12 +33,18 @@ type cloudWatchEventHandler struct {
...
 	appPublicURL       string
 	eventBusName       string
+	searchEndpoint     string
 }
 
 func (h *cloudWatchEventHandler) Handle(ctx context.Context, cloudWatchEvent events.CloudWatchEvent) error {
 	switch cloudWatchEvent.DetailType {
 	case "uid-requested":
-		uidStore := app.NewUidStore(h.dynamoClient, h.now)
+		searchClient, err := search.NewClient(h.cfg, h.searchEndpoint)
```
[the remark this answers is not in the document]
acsauk: https://opgtransform.atlassian.net/browse/MLPAB-1973

PR #1088, opened 2024-03-01
MLPAB-1791 Edit member permissions and enforce

hawx:


internal/page/donor/register.go
```diff
@@ -112,16 +111,7 @@ type Payer interface {
...
-	T(string) string
-	Count(messageID string, count int) string
-	FormatCount(messageID string, count int, data map[string]interface{}) string
-	ShowTranslationKeys() bool
-	SetShowTranslationKeys(s bool)
-	Possessive(s string) string
-	Concat([]string, string) string
-	FormatDate(date.TimeOrDate) string
-	FormatDateTime(time.Time) string
+	page.Localizer
```
hawx, line 114: we only have these as proxies so the mocks generate, may as well make that clear

acsauk (approved):
:shipit:

PR #1089, opened 2024-03-01
Refactor sesh package to remove gorilla/session imports everywhere

hawx:
I've been wanting to do this for a while now, the package design was always a pain especially in tests. This should work more "normally"


acsauk (approved):
Huge improvement, it was always a chore to work out the right mocks to set for these!

:shipit:

PR #1098, opened 2024-03-07
MLPAB-1976 Add data access policy for event-recevied

hawx:
# Purpose

Resolve errors related to indexing data in the events received lambda function.

Fixes MLPAB-1976

## Approach

- Refactor the creation of lambda function IAM roles out of the lambda function module
- Create lambda function IAM roles globally so that assigning permissions is easier
- Allow event-received lambda function to use opensearch


andrewpearce-digital (approved):

terraform/environment/opensearch.tf
```diff
@@ -73,3 +73,32 @@ resource "aws_opensearchserverless_access_policy" "lpas_collection_data_access_p
   ])
   provider = aws.eu_west_1
 }
+
+data "aws_iam_role" "event_received" {
```
andrewpearce-digital, line 77: This IAM role won't have been created yet, as it's in the lambda module.

The pattern has been to create iam roles globally, and pass them into region/modules, and I've not done that with the lambda module we use.

The reason for the pattern is to support adding permissions in just the way being done here.

I think the thing to do is to refactor the lambda module to require an iam role as an input variable, create the role globally and pass it in through the region module as we do with other iam roles.

andrewpearce-digital (changes requested):
see comment, we need to refactor things in order to support the change being made

andrewpearce-digital (approved):

acsauk (approved):
:shipit:

PR #1100, opened 2024-03-07
MLPAB-1991 - Fix AV Scan lambda function

andrewpearce-digital:
# Purpose

Use the correct log group (and improve initialisation times)

Fixes MLPAB-1991

## Approach

- set up logging config for av scan lambda function

## Learning


acsauk (approved):
:shipit:

PR #1102, opened 2024-03-08
MLPAB-1964 Use shared mock-onelogin image

hawx:


docker/docker-compose.yml
```diff
@@ -104,16 +104,15 @@ services:
       - "8081:8080"
 
   mock-onelogin:
-    build:
-      context: ..
-      dockerfile: docker/mock-onelogin/Dockerfile
+    image: 311462405659.dkr.ecr.eu-west-1.amazonaws.com/mock-onelogin:v0.62.0-beta.8
```
hawx, line 107: oops

docker/docker-compose.yml
```diff
@@ -104,16 +104,15 @@ services:
       - "8081:8080"
 
   mock-onelogin:
-    build:
-      context: ..
-      dockerfile: docker/mock-onelogin/Dockerfile
+    image: 311462405659.dkr.ecr.eu-west-1.amazonaws.com/mock-onelogin:v0.62.0
```
acsauk, line 107: Is the plan to manually keep the versioning up to date here? It would be nice if we could rely on a `latest` tag as it looks like we're already slightly out of sync

Screenshot 2024-03-08 at 10 37 25

hawx: 🤔 I guess that would work

acsauk (approved):
:shipit:

PR #1108, opened 2024-03-11
MLPAB-1998 - Event received lambda function fails

andrewpearce-digital:
# Purpose

Resolve error when invoking the lambda function

Fixes MLPAB-1998

## Approach

- Allow lambda to decrypt using aws/lambda key


acsauk (approved):
:shipit:

PR #1112, opened 2024-03-12
MLPAB-1541 Send codes to move online with paper-form-requested event

hawx:


internal/app/share_code_store.go
```diff
@@ -30,6 +31,10 @@ func (s *shareCodeStore) Get(ctx context.Context, actorType actor.Type, shareCod
 	}
 
 	err = s.dynamoClient.One(ctx, pk, sk, &data)
+	if data.Paper {
+		return actor.ShareCodeData{}, errors.New("paper codes not implemented yet")
```
hawx, line 35: I wanted a temporary check to prevent the codes being used until we actually build out the other side

acsauk (dismissed):
:shipit:

acsauk (approved):
:shipit:

PR #1119, opened 2024-03-14
MLPAB-1960 Add page to invite donor to supporter made LPA

hawx:


internal/page/supporter/donor_access.go
```diff
@@ -0,0 +1,99 @@
...
+				if err := shareCodeStore.PutDonor(r.Context(), shareCode, shareCodeData); err != nil {
+					return err
+				}
+
+				if err := notifyClient.SendEmail(r.Context(), data.Form.Email, notify.DonorAccessEmail{
+					ShareCode: shareCode,
+				}); err != nil {
+					return err
+				}
+
```
acsauk, line 69: There was a requirement on the ticket for updating the donor's email address if it differs from what was provided - did we end up deciding against this in refinement/review?

acsauk (commented):
One Q on updating the donor's email address but otherwise ready to merge.

internal/page/supporter/donor_access.go
hawx: ah yeah, I forgot about that actually being an update

acsauk (approved):
:shipit: although in a follow-up PR we should probably populate the email field in `your_details` for supporters as its currently write only (if that makes sense).

PR #1122, opened 2024-03-14
MLPAB-1999 - enable ECS deployment circuit breaker and rollback on failure

andrewpearce-digital:
# Purpose

When an ECS deployment fails, instead of trying until Terraform times out, break and rollback to a previous healthy deployment.

Notify the team that the circuit breaker has been thrown

Fixes MLPAB-1999

## Approach

- Enable circuit breaker, and set rollback to true for app
- Create Cloudwatch alert for failed deployments
- raise a pagerduty alert for the team when a deployment fails

## Learning

- https://aws.amazon.com/blogs/containers/announcing-amazon-ecs-deployment-circuit-breaker/
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_service#deployment_circuit_breaker


acsauk (approved):
:shipit:

PR #1123, opened 2024-03-14
MLPAB-1959: Add progress tracker to supporter view LPA

acsauk:
# Purpose

Fixes MLPAB-1959

After extracting the view logic into a function it was getting hard to test handlers that were using the progress tracker so I've made it its own type.

This change also enables us to hook into the existing donor fixtures in supporter fixtures by setting an extra flag in the query.


internal/actor/lpa_progress_tracker.go
```diff
@@ -0,0 +1,168 @@
...
+	)
+
+	if donor.CertificateProvider.FirstNames != "" {
+		certificateProviderSignedLabel = pt.Localizer.Format(
+			"certificateProviderHasDeclared",
+			map[string]interface{}{"CertificateProviderFullName": donor.CertificateProvider.FullName()},
+		)
+	} else {
+		if donor.IsOrganisationDonor() {
+			certificateProviderSignedLabel = pt.Localizer.T("theCertificateProviderHasDeclared")
```
hawx, line 42: I think it might be cleaner to have it structured like this (or similar)

```go
labels := map[string]string{
  "certificateProviderSigned": "yourCertificateProviderHasDeclared",
  // ...
}

if donor.IsOrganisationDonor() {
  labels = map[string]string{
    "certificateProviderSigned": "theCertificateProviderHasDeclared",
    // ...
  }
}

pt.Localizer.T(labels["certificateProviderSigned"])
```

hawx: although in the figma this is personalised as the certificate provider's name

hawx: which looks weird, I think it would be nicer to not personalise any of these and have it consistently "the donor", "the certificate provider", etc.

internal/actor/lpa_progress_tracker_test.go
```diff
@@ -0,0 +1,393 @@
...
+		Paid:                      ProgressTask{State: TaskNotStarted, Label: ""},
+		ConfirmedID:               ProgressTask{State: TaskNotStarted, Label: ""},
+		DonorSigned:               ProgressTask{State: TaskInProgress, Label: "You’ve signed your LPA"},
+		CertificateProviderSigned: ProgressTask{State: TaskNotStarted, Label: "Your certificate provider has provided their certificate"},
+		AttorneysSigned:           ProgressTask{State: TaskNotStarted, Label: "Your attorney has signed your LPA"},
+		LpaSubmitted:              ProgressTask{State: TaskNotStarted, Label: "We have received your LPA"},
+		StatutoryWaitingPeriod:    ProgressTask{State: TaskNotStarted, Label: "Your 4-week waiting period has started"},
+		LpaRegistered:             ProgressTask{State: TaskNotStarted, Label: "Your LPA has been registered"},
+	}
+	bundle, err := localize.NewBundle("../../lang/en.json")
```
hawx, line 28: I'd much prefer to have this mocked, or weblate changes could break these tests

internal/actor/lpa_progress_tracker.go
```diff
@@ -0,0 +1,168 @@
+package actor
```
hawx, line 1: I think this should live in `page` instead?

internal/actor/lpa_progress_tracker_test.go
acsauk: Ah yeah good point 👍

PR #1134, opened 2024-03-18
Update to use notification-banner layout in other places

hawx:
We probably should have made this a template sooner


acsauk (approved):
:shipit: definitely overdue!

PR #1137, opened 2024-03-19
MLPAB-1554 Generate UnmarshalText methods for enums

hawx:
I'm going to need this when reading from the lpa store, so may as well separate the mess


acsauk (approved):
:shipit:

PR #1138, opened 2024-03-19
MLPAB-2003: Add remove donor access to supporter LPA

acsauk:
# Purpose

Fixes MLPAB-2003

### DIFF
--- a/internal/app/donor_store.go
+++ b/internal/app/donor_store.go
@@ -335,6 +335,33 @@ func (s *donorStore) Delete(ctx context.Context) error {
 	return s.dynamoClient.DeleteKeys(ctx, keys)
 }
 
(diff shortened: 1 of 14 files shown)


internal/app/donor_store.go
```diff
@@ -335,6 +335,32 @@ func (s *donorStore) Delete(ctx context.Context) error {
...
+	data, err := page.SessionDataFromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	if data.OrganisationID == "" {
+		return errors.New("donorStore.DeleteLink requires OrganisationID")
+	}
+
+	if data.OrganisationID != shareCodeData.SessionID {
```
acsauk, line 348: Slight overkill but as we had the data handy thought we may as well be sure the remover has the rights to do it.

hawx (approved):

PR #1140, opened 2024-03-20
MLPAB-1554 Send certificate provider prompt when LPA ready

hawx:


internal/actor/actoruid/uid.go
```diff
@@ -8,6 +8,9 @@ import (
 
 const prefix = "urn:opg:poas:makeregister:users:"
 
+// Service is used when the application does an action.
+var Service = UID{value: [16]byte{0, 0, 0, 0, 0, 0, 64}}
```
hawx, line 12: I didn't want to use the null UUID here, so settled for setting the version to 4. We can always change this later if/when we decide how to represent it. But it shouldn't really matter too much just for getting an LPA as that isn't recorded anywhere.

cmd/event-received/main.go
```diff
@@ -31,6 +31,12 @@ type uidEvent struct {
 	UID string `json:"uid"`
 }
 
+type changeEvent struct {
```
acsauk, line 34: Don't think this is being used anywhere at the moment?

acsauk (commented):
One unused type to drop then ready to merge

cmd/event-received/main.go
hawx: ah yeah, left over from when I had the other event

acsauk (approved):
:shipit:

PR #1155, opened 2024-03-26
MLPAB-1535: Vary certificate provider contact number label by donor Channel

acsauk:
# Purpose

Fixes MLPAB-1535


.github/workflows/workflow_pr.yml
```diff
@@ -97,7 +97,7 @@ jobs:
     with:
       run_against_image: true
       tag: ${{ needs.create_tags.outputs.version_tag }}
-      skip: ${{ contains(fromJSON('["weblate-pr"]'), github.head_ref) }}
+      skip: true
```
acsauk, line 100: As per slack, cypress is bugging out at the moment and failing for unknown reasons on this branch and fresh branches pushed from main

internal/actor/certificate_provider_provided.go
```diff
@@ -31,6 +31,8 @@ type CertificateProviderProvidedDetails struct {
 	Tasks CertificateProviderTasks
 	// ContactLanguagePreference is the language the certificate provider prefers to receive notifications in
 	ContactLanguagePreference localize.Lang
+	// DonorChannel is the way the donor is applying for the LPA (online or paper)
+	DonorChannel Channel
```
acsauk, line 35: I was on the fence between adding this to the CP vs persisting a paper LPA in dynamo - having a shadow of the LPA store LPA felt like it could lead to some data inconsistencies but this means in a future PR we'll need to point all LPA gets for CPs (and eventually attorneys with a paper donor) to the lpa store rather than dynamo.

hawx: maybe this ticket should be done after the one I'm currently doing, because that will be the case after I'm done

.github/workflows/workflow_pr.yml
hawx: I thought only 6 tests were failing, it would be better not to skip the tests that do work

internal/page/certificateprovider/confirm_your_details.go
```diff
@@ -47,6 +48,11 @@ func ConfirmYourDetails(tmpl template.Template, donorStore DonorStore, certifica
 			App:                 appData,
 			CertificateProvider: certificateProvider,
 			Donor:               donor,
+			PhoneNumberLabel:    "mobileNumber",
+		}
+
+		if certificateProvider.DonorChannel.IsPaper() {
```
hawx, line 54: I think there was a simpler implementation than passing this everywhere, we know the donor is on paper if `donorStore.GetAny(ctx)` returns a `dynamo.NotFoundError{}`. That would then also save us having to rework as much later

acsauk:
Closing in favour of #1167

PR #1166, opened 2024-04-04
MLPAB-1223 Add missing data to fixtures sent to lpa-store

hawx:


acsauk (approved):
:shipit:

internal/page/fixtures/attorney.go
```diff
@@ -189,6 +174,7 @@ func Attorney(
 
 		donorDetails.AttorneyDecisions = actor.AttorneyDecisions{How: actor.JointlyAndSeverally}
 		donorDetails.ReplacementAttorneyDecisions = actor.AttorneyDecisions{How: actor.JointlyAndSeverally}
+		donorDetails.HowShouldReplacementAttorneysStepIn = actor.ReplacementAttorneysStepInWhenAllCanNoLongerAct
```
hawx, line 177: TODO: check this should be required, if so add to the others

internal/lpastore/resolving_service.go
```diff
@@ -37,7 +37,7 @@ func (s *ResolvingService) Get(ctx context.Context) (*Lpa, error) {
 
 	lpa, err := s.client.Lpa(ctx, donor.LpaUID)
 	if errors.Is(err, ErrNotFound) {
-		lpa = &Lpa{}
+		lpa = donorProvidedDetailsToLpa(donor)
```
hawx, line 40: I wanted us to not do this, but we need to so that we can show data when the certificate provider has got the first email but the donor hasn't signed

PR #1167, opened 2024-04-04
MLPAB-1535: Vary certificate provider phone field label for paper donors

acsauk:
# Purpose

Fixes MLPAB-1535

### DIFF
--- a/internal/page/fixtures/attorney.go
+++ b/internal/page/fixtures/attorney.go
@@ -280,7 +280,7 @@ func Attorney(
 		if err := donorStore.Put(donorCtx, donorDetails); err != nil {
 			return err
 		}
-		if donorDetails.LpaUID != "" {
+		if !donorDetails.SignedAt.IsZero() && donorDetails.LpaUID != "" {
 			if err := lpaStoreClient.SendLpa(donorCtx, donorDetails); err != nil {
 				return err
 			}
--- a/internal/page/fixtures/donor.go
+++ b/internal/page/fixtures/donor.go
@@ -116,7 +116,7 @@ func Donor(
 		if err := donorStore.Put(donorCtx, donorDetails); err != nil {
 			return err
 		}
-		if donorDetails.LpaUID != "" {
+		if !donorDetails.SignedAt.IsZero() && donorDetails.LpaUID != "" {
 			if err := lpaStoreClient.SendLpa(donorCtx, donorDetails); err != nil {
 				return err
 			}
--- a/internal/page/fixtures/supporter.go
+++ b/internal/page/fixtures/supporter.go
@@ -115,9 +115,6 @@ func Supporter(
 				if err := donorStore.Put(donorCtx, donor); err != nil {
 					return err
 				}
-				if err := lpaStoreClient.SendLpa(donorCtx, donor); err != nil {
-					return err
-				}
 
 				shareCodeData := actor.ShareCodeData{
 					SessionID:    org.ID,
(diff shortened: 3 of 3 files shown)


cypress/e2e/certificate-provider/confirm-your-details.cy.js
```diff
@@ -1,7 +1,7 @@
 describe('Confirm your details', () => {
   describe('shows details for', () => {
     it('lay certificate providers', () => {
-      cy.visit('/fixtures/certificate-provider?redirect=/enter-date-of-birth');
+      cy.visit('/fixtures/certificate-provider?redirect=/enter-date-of-birth&sendLpaToLpaStore=1');
```
hawx, line 4: if you remove this the test still passes, so does it need to be sent?

internal/lpastore/lpa.go
```diff
@@ -246,6 +246,7 @@ type Lpa struct {
 	Submitted                                  bool
 	Paid                                       bool
 	IsOrganisationDonor                        bool
+	IsPaperDonor                               bool
```
hawx, line 249: Would it be better to add a `Channel` to `Donor`, then set that? We could then set it to the correct option when a donor is created in dynamo/read from lpastore

internal/page/certificateprovider/confirm_your_details.go
```diff
@@ -48,6 +49,11 @@ func ConfirmYourDetails(tmpl template.Template, lpaStoreResolvingService LpaStor
 			App:                 appData,
 			CertificateProvider: certificateProvider,
 			Lpa:                 lpa,
+			PhoneNumberLabel:    "mobileNumber",
+		}
+
+		if lpa.IsPaperDonor {
+			data.PhoneNumberLabel = "contactNumber"
```
hawx, line 56: There is already some label switching in the template. I think it would be good to either have all the label logic in here, or all in the template (and then maybe we decide that we will stick to that option going forward?)

cypress/e2e/certificate-provider/confirm-your-details.cy.js
acsauk: Hmmm interesting - I was getting the details back from the LPA store rather than the donor-entered details but this was before I'd merged in main a couple of times. All looks fine, as you pointed out, so I'll drop this.

internal/page/certificateprovider/confirm_your_details.go
hawx: :+1: that's fine logic for me

web/template/certificateprovider/confirm_your_details.gohtml
```diff
@@ -42,11 +42,9 @@
       <h2 class="govuk-heading-m govuk-!-margin-top-8">{{ tr .App "detailsTheDonorHasGivenAboutYou" }}</h2>
 
       {{ $detailsContent := "whatToDoIfAnyDetailsAreIncorrectCertificateProviderContentLay" }}
-      {{ $addressLabel := "address" }}
 
       {{ if .Lpa.CertificateProvider.Relationship.IsProfessionally }}
         {{ $detailsContent = "whatToDoIfAnyDetailsAreIncorrectCertificateProviderContentProfessional" }}
```
hawx, line 47: I know this isn't a label, but I think we should do it for all switched content regardless of what it is

hawx: I do like the idea of moving logic out of the templates so it can all be tested

hawx (approved):

PR #1172, opened 2024-04-08
Only send to lpa-store in fixtures if signed

hawx:
For some reason I forgot to put this check on the others...


acsauk (approved):
:shipit:

PR #1180, opened 2024-04-10
MLPAB-1546: Certificate provider moving from paper to digital

acsauk:
# Purpose

Storing email on CP is likely a temporary step until we decide how we want to handle immaterial changes to an LPA (as the LPA may have already been sent to the LPA store when the CP first logs in). This change at least allows a paper CP to move to online and receive emails.

Fixes MLPAB-1546


internal/page/donor/how_would_certificate_provider_prefer_to_carry_out_their_role.go
```diff
@@ -58,9 +58,14 @@ type howWouldCertificateProviderPreferToCarryOutTheirRoleForm struct {
 func readHowWouldCertificateProviderPreferToCarryOutTheirRole(r *http.Request) *howWouldCertificateProviderPreferToCarryOutTheirRoleForm {
 	carryOutBy, err := actor.ParseCertificateProviderCarryOutBy(page.PostFormString(r, "carry-out-by"))
 
+	email := page.PostFormString(r, "email")
+	if carryOutBy.IsPaper() {
+		email = ""
+	}
```
acsauk, line 64: I ended up adding this in as we [validate](https://github.com/ministryofjustice/opg-data-lpa-store/blob/main/lambda/create/validate.go#L25) on this field not being present in the lpa store for paper CPs.

internal/actor/certificate_provider_provided.go
```diff
@@ -31,6 +31,8 @@ type CertificateProviderProvidedDetails struct {
 	Tasks CertificateProviderTasks
 	// ContactLanguagePreference is the language the certificate provider prefers to receive notifications in
 	ContactLanguagePreference localize.Lang
+	// Email is the email address returned from OneLogin when the certificate provider logged in
+	Email string
```
hawx, line 35: was this meant to be MLPAB-1546, and so this needs sending to the lpa-store?

acsauk: My bad - I'd looked at the wrong column when grabbing the ticket number. Updated now.

internal/lpastore/update.go
```diff
@@ -36,7 +36,7 @@ func (c *Client) sendUpdate(ctx context.Context, lpaUID string, actorUID actorui
 	return c.do(ctx, actorUID, req, nil)
 }
 
-func (c *Client) SendCertificateProvider(ctx context.Context, lpaUID string, certificateProvider *actor.CertificateProviderProvidedDetails) error {
+func (c *Client) SendCertificateProvider(ctx context.Context, lpaUID string, certificateProvider *actor.CertificateProviderProvidedDetails, lpa *Lpa) error {
```
hawx, line 39: you could make the signature match `SendAttorney` and remove `lpaUID` here

internal/lpastore/update.go
```diff
@@ -46,27 +46,35 @@ func (c *Client) SendCertificateProvider(ctx context.Context, lpaUID string, cer
...
 	if certificateProvider.HomeAddress.Country != "" {
-		body.Changes = append(body.Changes, updateRequestChange{Key: "/certificateProvider/address/country", New: certificateProvider.HomeAddress.Country})
+		body.Changes = append(body.Changes, updateRequestChange{Key: "/certificateProvider/address/country", New: certificateProvider.HomeAddress.Country, Old: lpa.CertificateProvider.Address.Country})
+	}
+
+	if certificateProvider.Email != "" {
+		body.Changes = append(body.Changes, updateRequestChange{Key: "/certificateProvider/email", New: certificateProvider.Email, Old: lpa.CertificateProvider.Email})
+	}
+
+	if lpa.CertificateProvider.Channel != actor.ChannelOnline {
```
hawx, line 76: I know we only have 2 channels, but I'd prefer `== channel.Paper` here as that is what you are saying it is changing from

hawx (approved):

PR #1184, opened 2024-04-12
MLPAB-2036 Use lpa-store to work out when attorneys have signed

hawx:


package.json
```diff
@@ -10,7 +10,7 @@
         "cypress:open": "node_modules/.bin/cypress open",
         "cypress:run": "node_modules/.bin/cypress run -vvv",
         "cypress:parallel": "cypress-parallel -s cypress:run -t 6 --spec cypress/e2e/**/*.cy.js cypress/smoke/*.cy.js",
-        "cypress:parallel-with-specs": "cypress-parallel -s cypress:run -t 4",
+        "cypress:parallel-with-specs": "cypress-parallel -s cypress:run -t 2",
```
hawx, line 13: I kept getting the first test of one thread failing until I reduced it to this. I'm going to lump the investigation around why into the existing Cypress ticket, because I'm not sure if that is going to be the same issue

acsauk (approved):
:shipit:

PR #1189, opened 2024-04-17
MLPAB-2067 - Only deploy OpenSearch Ingestion Pipeline when needed

andrewpearce-digital:
# Purpose

Save build time by not deploying ingestion pipelines when they aren't needed. This will not need to be per environment, but instead will be enabled when dependant work is completed.

Fixes MLPAB-2067

## Approach

- create a flag
- create resources if true
- re-enable app search indexing


acsauk (approved):
:shipit:

acsauk (approved):
:shipit:

PR #1190, opened 2024-04-18
MLPAB-2035 Use lpa-store for registered status

hawx:


internal/lpastore/update.go
```diff
@@ -36,6 +36,14 @@ func (c *Client) sendUpdate(ctx context.Context, lpaUID string, actorUID actorui
 	return c.do(ctx, actorUID, req, nil)
 }
 
+func (c *Client) SendRegister(ctx context.Context, lpaUID string) error {
```
hawx, line 39: I don't really like this, but the alternative was to make `SendUpdate` or `Do` public, which I think would be worse

internal/lpastore/resolving_service.go
```diff
@@ -61,11 +64,8 @@ func (s *ResolvingService) Get(ctx context.Context) (*Lpa, error) {
 		// copy the relationship as it isn't stored in the lpastore.
 		lpa.CertificateProvider.Relationship = donor.CertificateProvider.Relationship
 
-		// TODO: eventually we'll need to remove the RegisteredAt field as mlpa
-		// won't be tracking that data, then we'll need to figure out how to expose
-		// the data for testing
-		if !donor.RegisteredAt.IsZero() {
-			lpa.RegisteredAt = date.FromTime(donor.RegisteredAt)
+		if lpa.WithdrawnAt.IsZero() {
```
hawx, line 67: I'm going to add another ticket for this one 😄

internal/lpastore/update.go
acsauk: Fair compromise 👍

acsauk (approved):
:shipit:

PR #1193, opened 2024-04-18
MLPAB-2015 - Fix failed deployment alarms

andrewpearce-digital:
# Purpose

The Eventbridge rule correctly matches the failed deployment, but fails to process it to a log group. The metric alarm then doesn't raise the alert to the team.

Fixes MLPAB-2015

## Approach

- fix event matching in rule
- put event logs in `/aws/events/` path
- create insights query for for log group
- ~fix raise pagerduty alert when deployment failed metric alarm raised~ (will be fixed in org-infra)

## Learning

- https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-event-patterns.html
- https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_service_deployment_events.html


acsauk (approved):
:shipit:

PR #1198, opened 2024-04-23
MLPAB-2034 Dynamo keys

hawx:
I'd recommend reading [internal/dynamo/keys.go](https://github.com/ministryofjustice/opg-modernising-lpa/pull/1198/files#diff-0cbb67583f71bb65d5efb143161ec805eb49c9de0719dd5860a6d0344b444a2d) and [internal/dynamo/complex_keys.go](https://github.com/ministryofjustice/opg-modernising-lpa/pull/1198/files#diff-4a6072316e7cf0afa3680a57370a70f21fa5fbef4fd30d24bd5a83442bba0684) first. Most of the other changes are me applying these functions everywhere


internal/dynamo/complex_keys.go
```diff
@@ -0,0 +1,248 @@
...
+	if err := json.Unmarshal(text, &skeys); err != nil {
+		return err
+	}
+
+	return k.read(skeys.PK, skeys.SK)
+}
+
+type LpaOwnerKeyType struct{ sk SK }
+
+func LpaOwnerKey(sk interface {
```
hawx, line 61: this was the best way I could think of dealing with the situations where we need to allow multiple keys for the same field, it is a bit clunky but mainly a problem that you only see in tests

internal/dynamo/keys.go
```diff
@@ -2,107 +2,239 @@ package dynamo
...
-func LpaKey(s string) string {
-	return "LPA#" + s
+func LpaKey(s string) LpaKeyType {
+	return LpaKeyType(lpaPrefix + "#" + s)
 }
 
+type DonorKeyType string
+
+func (t DonorKeyType) SK() string { return string(t) }
+func (t DonorKeyType) lpaOwner()  {}
```
acsauk, line 88: I get what this is doing but it may be confusing for future maintainers. I'm not sure if adding a comment here and on `OrganisationKeyType` or in [complex_keys](https://github.com/ministryofjustice/opg-modernising-lpa/pull/1198/files#diff-4a6072316e7cf0afa3680a57370a70f21fa5fbef4fd30d24bd5a83442bba0684R61) makes more sense but some pointer to its purpose may be helpful?

acsauk (commented):
Adding both safety and context around our key schemas, great PR 👏

One small request for some clarity on lpa owners and then good to go.

acsauk (approved):
:shipit:

PR #1201, opened 2024-04-26
MLPAB-1777 Show when notices of intent to register have been sent

hawx:


cmd/event-received/makeregister_event_handler.go
```diff
@@ -0,0 +1,48 @@
...
+		if err != nil {
+			return err
+		}
+
+		uidClient := factory.UidClient()
+
+		return handleUidRequested(ctx, uidStore, uidClient, cloudWatchEvent)
+
+	default:
+		return fmt.Errorf("unknown cloudwatch event")
```
acsauk, line 28: Should this be updated to `makeregister` event now we're splitting by service?

internal/page/lpa_progress_tracker.go
```diff
@@ -45,16 +64,16 @@ func (pt ProgressTracker) Progress(lpa *lpastore.Lpa) Progress {
...
 			"lpaRegistered":             pt.Localizer.T("theLpaHasBeenRegistered"),
 		}
 	} else {
 		labels = map[string]string{
-			"paid":                   "",
-			"confirmedID":            "",
 			"donorSigned":            pt.Localizer.T("youveSignedYourLpa"),
 			"attorneysSigned":        pt.Localizer.Count("attorneysHaveDeclared", len(lpa.Attorneys.Attorneys)),
 			"lpaSubmitted":           pt.Localizer.T("weHaveReceivedYourLpa"),
+			"noticesOfIntentSent":    "weSentAnEmailYourLpaIsReadyToRegister",
```
acsauk, line 76: Did we come to a decision yet on status changes vs notification sent for showing this?

acsauk (commented):
A couple of questions 💬

internal/page/lpa_progress_tracker.go
hawx: Sirius will change the status in lpa-store, that will then cause an event for us? (Is that what you mean?)

acsauk (approved):
:shipit:

PR #1207, opened 2024-04-29
MLPAB-1779 Update status colours and wording

hawx:


acsauk (commented):
Unless there were some separate discussions I think a couple of the colours need updating to match the figma/sheet

acsauk (approved):
:shipit:

PR #1208, opened 2024-04-30
MLPAB-1616 Update navigation for the attorney journey

hawx:


internal/templatefn/fn.go
```diff
@@ -42,6 +42,7 @@ func All(globals *Globals) map[string]any {
 		"isEnglish":          isEnglish,
 		"isWelsh":            isWelsh,
 		"input":              input,
+		"button":             button,
```
hawx, line 45: I'm calling it button because we might want to do a `{{ template "button" (button ...) }}` later

web/template/layout/buttons.gohtml
```diff
@@ -0,0 +1,22 @@
+{{ define "buttons" }}
```
acsauk, line 1: Nice, this drop looooads of boilerplate

acsauk (approved):
:shipit:

PR #1210, opened 2024-05-01
MLPAB-2088: Conditionally send Channel and Email for TRUST_CORPORATION_SIGN

acsauk:
# Purpose

Fixes MLPAB-2088

### DIFF
--- a/internal/lpastore/update.go
+++ b/internal/lpastore/update.go
@@ -130,6 +130,15 @@ func (c *Client) SendAttorney(ctx context.Context, lpa *Lpa, attorney *actor.Att
 
 	if attorney.IsTrustCorporation {
 		body.Type = "TRUST_CORPORATION_SIGN"
+		lpaTrustCorp := lpa.Attorneys.TrustCorporation
+
+		if lpaTrustCorp.Email != attorney.Email {
+			body.Changes = append(body.Changes, updateRequestChange{Key: attorneyKey + "/email", New: attorney.Email, Old: lpaTrustCorp.Email})
+		}
+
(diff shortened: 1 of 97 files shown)


internal/lpastore/update.go
```diff
@@ -130,6 +130,15 @@ func (c *Client) SendAttorney(ctx context.Context, lpa *Lpa, attorney *actor.Att
 
 	if attorney.IsTrustCorporation {
 		body.Type = "TRUST_CORPORATION_SIGN"
+		lpaTrustCorp := lpa.Attorneys.TrustCorporation
```
hawx, line 133: you'll need to check if it's a replacement, then use `lpa.ReplacementAttorneys.TrustCorporation` instead

web/template/layout/attorney-summary.gohtml
```diff
@@ -41,19 +41,21 @@
...
-              <dd class="govuk-summary-list__actions">
-                <a class="govuk-link govuk-link--no-visited-state" href="{{ link .App .Link.TrustCorporation }}#f-email">
-                  {{ trHtml .App "changeCompanyEmailLink" }}
-                </a>
-              </dd>
-            {{ end }}
-          </div>
-          
+
+          {{ if .TrustCorporation.Email }}
```
hawx, line 45: I think we should always show the row, then they can use Change to set one from this page

acsauk: I was following the pattern for attorneys where we hide the row but I agree it would be useful to be able to change this. I'll add in and raise with Beth at next dev review.

hawx (approved):

PR #1219, opened 2024-05-07
MLPAB-2059 Index donor names and alter queries to work with that

hawx:


internal/page/fixtures/supporter.go
```diff
@@ -303,11 +304,23 @@ func Supporter(
...
 			Page:     1,
 			PageSize: 1,
-		}); resp != nil && len(resp.Keys) > 0 {
+		})
+		if err != nil {
+			log.Println(err)
+		}
+
+		if count > 10 {
+			return
```
hawx, line 319: this will stop the infinite loop of checking

internal/search/client.go
```diff
@@ -84,18 +88,26 @@ func NewClient(cfg aws.Config, endpoint, indexName string, indexingEnabled bool)
...
 	}
 
-	return &Client{indices: svc.Indices, svc: svc, endpoint: endpoint, indexName: indexName, indexingEnabled: indexingEnabled}, nil
+	hash := sha256.New()
+	if err := json.NewEncoder(hash).Encode(indexDefinition); err != nil {
+		return nil, err
+	}
+	sum := hash.Sum(nil)
+
+	return &Client{indices: svc.Indices, svc: svc, endpoint: endpoint, indexName: fmt.Sprintf("%s_%x", indexName, sum[:8]), indexingEnabled: indexingEnabled}, nil
```
hawx, line 97: this adds a hash of the config to the index name, which will prevent problems when it changes but obviously will need the later work to make re-indexing/removal of old ones to be a finished solution

andrewpearce-digital (approved):

PR #1220, opened 2024-05-07
Add LPA channel to data sent to LPA Store

gregtyler:
# Purpose

This will allow us differentiate between LPAs created through MRLPA and through scanning.

## Approach

It's always "online".

## Learning

Plenty: it's my first proper contribution to this repo! But it was all very straightforward, thank you


acsauk (commented):
All looks good but could we get `Channel` added to the [lpa response](https://github.com/ministryofjustice/opg-modernising-lpa/blob/0bbcd588607264f6d7d5b136d5acb2d16acca170/internal/lpastore/lpa.go#L301) from the store too? We will need it for the MRLPA side of CP opt-out I'm currently working on to account for paper CPs and where we send comms to.

gregtyler:
@acsauk Yep, but the Pact rituals require that I add it in LPA Store after this PR, then come back and add it here.

acsauk (approved):
Ah ha, gotta love this dance 😂 

:shipit:

PR #1222, opened 2024-05-08
Add `channel` to LPAs returned from LPA Store

gregtyler:
# Purpose

To be used in CP opt-out.

Fixes MLPAB-1535

## Approach

The property is already in the LPA Store, so this just adds it to structs, testing and Pact contract.


acsauk (approved):
:shipit:

PR #1226, opened 2024-05-10
MLPAB-2092 - Deploy a zip based lambda function for S3 antivirus scanning

andrewpearce-digital:
# Purpose

Improve performance by using zip packages and lambda layers to reduce initialisation time during execution.

Fixes MLPAB-2092

## Approach

- Create a lambda layer
- Create a zip package function
- Use new function for object scanning
- make sure this works for ur deployments too (it doesn't run create tags) 

## Learning

- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_function
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lambda_layer_version


acsauk (approved):
:shipit:

PR #1228, opened 2024-05-10
MLPAB-2062 Log email address when send fails

hawx:
Also logging the trace_id with everything so we can tie things together, we just need to always use the `WhateverContext` logging methods


internal/telemetry/telemetry.go
```diff
@@ -75,3 +76,33 @@ func WrapHandler(handler http.Handler) http.HandlerFunc {
...
+	handler slog.Handler
+}
+
+func NewSlogHandler(h slog.Handler) slog.Handler {
+	return &SlogHandler{handler: h}
+}
+
+func (h *SlogHandler) Enabled(ctx context.Context, level slog.Level) bool {
+	return h.handler.Enabled(ctx, level)
+}
```
acsauk, line 90: Is this function and `WithGroup()` here to satisfy an interface as they don't appear to be doing anything or be in use?

acsauk (approved):
:shipit:

PR #1230, opened 2024-05-14
MLPAB-1835 Remove ability to skip address entry

hawx:


cypress/e2e/donor/choose-attorneys-task.cy.js
```diff
@@ -1,4 +1,4 @@
-import { TestEmail } from "../../support/e2e";
+import { AddressFormAssertions, TestEmail } from "../../support/e2e";
```
hawx, line 1: I think a good time to unskip these will be when we get to the ticket that is for changing some of the logic

acsauk (approved):
:shipit:

PR #1233, opened 2024-05-15
MLPAB-2098: Add correspondent to fixture progress, protect against opensearch not being ready

acsauk:
# Purpose

Fixes MLPAB-2089

## Learning

[depends_on](https://stackoverflow.com/a/74449969/7648583) has more options than I knew 🤷


docker/docker-compose.yml
```diff
@@ -93,6 +94,12 @@ services:
         hard: -1
     volumes:
       - data01:/usr/share/opensearch/data
+#    To ensure opensearch is ready for localstack-init commands
```
acsauk, line 97: I was having issues when freshly bringing up the app when opensearch wasn't ready so I've added this to make sure it's ready for the indices to be created.

internal/page/fixtures/supporter.go
```diff
@@ -312,6 +312,7 @@ func waitForLPAIndex(searchClient *search.Client, organisationCtx context.Contex
 			PageSize: 1,
 		})
 		if err != nil {
+			log.Println("error waiting for LPA Index")
```
acsauk, line 315: So its immediately obvious what the stream of console errors are when this happens

docker/docker-compose.yml
hawx: you can probably remove the health check from localstack then, that will be left over from when we were using the built-in opensearch

internal/page/fixtures/donor.go
```diff
@@ -303,6 +304,19 @@ func updateLPAProgress(
 		donorDetails.Tasks.PeopleToNotify = actor.TaskCompleted
 	}
 
+	if data.Progress == slices.Index(progressValues, "addCorrespondent") {
```
hawx, line 307: do `>=` for these

internal/page/fixtures/donor.go
```diff
@@ -303,6 +304,19 @@ func updateLPAProgress(
...
+		donorDetails.AddCorrespondent = form.Yes
+		donorDetails.Correspondent = makeCorrespondent(Name{
+			Firstnames: "Jonathan",
+			Lastname:   "Ashfurlong",
+		})
+
+		donorDetails.Tasks.AddCorrespondent = actor.TaskCompleted
+
+		donorDetails.CheckedAt = time.Time{}
+		donorDetails.Tasks.CheckYourLpa = actor.TaskNotStarted
```
hawx, line 317: these should be the default values

internal/page/fixtures/donor.go
```diff
@@ -45,6 +45,7 @@ var progressValues = []string{
 	"chooseYourCertificateProvider",
 	"peopleToNotifyAboutYourLpa",
 	"checkAndSendToYourCertificateProvider",
+	"addCorrespondent",
```
hawx, line 48: you'll want this moved up 1 space

internal/page/fixtures/supporter.go
hawx: or could have the one below be `log.Println("error waiting for LPA index:", err)`

internal/page/fixtures/donor.go
hawx: also add the option to the template?

internal/page/fixtures/donor.go
acsauk: My thinking here was to make this an explicit choice given that in the future all comms will be routed to the correspondent rather than the donor which may get confusing when setting up fixtures.

internal/page/fixtures/donor.go
acsauk: Ah yeah, side effect of the placement of this section in the progress list. I'll remove.

internal/page/fixtures/donor.go
hawx: yeah, if we are doing special cases it would be better under a different option and keep `progress` as a pretty straightforward "fills in up to this step"

hawx (approved):

PR #1242, opened 2024-05-20
MLPAB-2126 Re-order attorney sign handlers to work when failure

hawx:


internal/page/attorney/sign_test.go
```diff
@@ -305,6 +337,7 @@ func TestPostSign(t *testing.T) {
 		appData         page.AppData
 		form            url.Values
 		lpa             *lpastore.Lpa
+		sendsToLpaStore bool
```
acsauk, line 340: This looks like its unused at the monent.

acsauk (commented):
Splitting the sign handler by attorney type is clearer now 👍 

One leftover testcase prop that probably needs to be removed then good to go.

internal/page/attorney/sign_test.go
hawx: :rage1: I thought I'd removed that

acsauk (approved):
:shipit:

PR #1244, opened 2024-05-20
MLPAB-2125: Use transactions when creating attorneys

acsauk:
# Purpose

Fixes MLPAB-2125


internal/app/attorney_store.go
```diff
@@ -29,24 +29,37 @@ func (s *attorneyStore) Create(ctx context.Context, lpaOwnerKey dynamo.LpaOwnerK
...
 		ActorType: actor.TypeAttorney,
 		UpdatedAt: s.now(),
-	}); err != nil {
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	transaction := dynamo.NewTransaction().
+		Put(marshalledAttorney).
```
hawx, line 57: I like this, my only suggestion would be to push the `MarhsalMap` calls in to `Put`. You could use a pattern like https://go.dev/blog/errors-are-values to hold onto the errors, then return in `WriteTransaction`

internal/page/fixtures/attorney.go
```diff
@@ -384,3 +481,27 @@ func Attorney(
...
+	if isReplacement {
+		attorneyType = actor.TypeReplacementAttorney
+	}
+
+	err := shareCodeStore.Put(ctx, attorneyType, shareCode, shareCodeData)
+	if err != nil {
+		return nil, err
+	}
+
+	return attorneyStore.Create(ctx, shareCodeData, email)
```
hawx, line 506: This is fine, it's the price we pay. I'd rather have the actual code and fixtures a mess. It's only for setting up test data so we shouldn't worry too much about it

internal/app/attorney_store.go
acsauk: Ah cool, that's precisely what I was trying to avoid so thanks for the link 👍

acsauk: (and this will make testing so much easier)

hawx (approved):

PR #1247, opened 2024-05-21
MLPAB-2029 - create an account level opensearch serverless collection

andrewpearce-digital:
# Purpose

Share a single collection to reduce costs

Fixes MLPAB-2029

## Approach

- Create an account level collection

## Learning

### DIFF
--- a/terraform/account/opensearch.tf
+++ b/terraform/account/opensearch.tf
@@ -1,0 +1,207 @@
+data "aws_vpc_endpoint" "opensearch" {
+  tags = {
+    Name = "opensearch-eu-west-1"
+  }
+  provider = aws.eu_west_1
+}
+
+data "aws_kms_alias" "opensearch" {
+  name     = "alias/${local.default_tags.application}-opensearch-encryption-key"
+  provider = aws.eu_west_1
+}
+
+resource "aws_opensearchserverless_security_policy" "lpas_collection_encryption_policy" {
+  name        = "policy-${local.account_name}"
+  type        = "encryption"
+  description = "encryption policy for collection"
+  policy = jsonencode({
+    Rules = [
+      {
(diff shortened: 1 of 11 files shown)


terraform/account/opensearch.tf
```diff
@@ -0,0 +1,197 @@
...
+      ],
+      Principal = [
+        "arn:aws:iam::${data.aws_caller_identity.global.account_id}:role/operator"
+      ]
+    }
+  ])
+  provider = aws.eu_west_1
+}
+
+resource "aws_opensearchserverless_access_policy" "team_breakglass_access" {
```
acsauk, line 91: Would we always want this available on production? Or have we decided opensearch is always off-limits for live data?

acsauk: Oh ignore me, I've just seen the ordering is different here for `count` 🙄

terraform/account/terraform.tfvars.json
```diff
@@ -6,23 +6,26 @@
...
     },
     "production": {
       "account_id": "313879017102",
       "account_name": "production",
       "is_production": true,
       "regions": [
         "eu-west-1"
-      ]
+      ],
+      "pagerduty_service_name": "OPG Modernising LPA Non-Production"
```
acsauk, line 28: Should this be:

```suggestion
      "pagerduty_service_name": "OPG Modernising LPA Production"
```

andrewpearce-digital: We don't currently have a production pagerduty service to alert to, so this isn't a desired name but the name of an existing resource

terraform/account/opensearch.tf
andrewpearce-digital: Breakglass is an alarmed role (using it raises an alarm) for use in emergencies.

PR #1248, opened 2024-05-21
MLPAB-2128 Make provide certificate tolerate failures

hawx:


internal/notify/client.go
```diff
@@ -208,3 +246,12 @@ func (c *Client) do(req *http.Request) (response, error) {
 
 	return r, nil
 }
+
+func (c *Client) makeReference(to string, email Email) string {
```
hawx, line 250: I don't think there is any email we'd want to send to the same person multiple times, right?

internal/notify/client.go
```diff
@@ -165,10 +185,33 @@ func (c *Client) SendActorSMS(ctx context.Context, to, lpaUID string, sms SMS) e
...
+	}
+
+	resp, err := c.do(req)
+	if err != nil {
+		return false, err
+	}
+
+	for _, notification := range resp.Notifications {
+		if (notification.Status == "sending" || notification.Status == "delivered") &&
+			notification.CreatedAt.After(c.now().Add(-allowResendAfter)) {
```
hawx, line 201: I'm kind of scared of setting a blanket policy on preventing resends (which would be 7 days based on what you can query in notify), so I'm choosing a 10 minute window as that will prevent problems with people resubmitting the form

acsauk: Sensible 👍

internal/page/certificateprovider/provide_certificate.go
```diff
@@ -67,12 +71,13 @@ func ProvideCertificate(
...
-				}
 
-				if err := lpaStoreClient.SendCertificateProvider(r.Context(), certificateProvider, lpa); err != nil {
-					return err
+				if lpa.CertificateProvider.SignedAt.IsZero() {
+					if err := lpaStoreClient.SendCertificateProvider(r.Context(), certificateProvider, lpa); err != nil {
+						return err
+					}
+				} else {
+					certificateProvider.Certificate.Agreed = lpa.CertificateProvider.SignedAt
```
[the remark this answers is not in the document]
acsauk: Maybe that's a sign we align our term for signing to be `SignedAt`?

acsauk (approved):
:shipit: 

The comment on aligning terms probably applies in a number of places so could be good to look at where we differ from the LPA store way of naming things to see if it can be aligned.

internal/page/certificateprovider/provide_certificate.go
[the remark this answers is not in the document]
hawx: yes, and I'd love to remove the `.Certificate` wrapper

PR #1249, opened 2024-05-21
MLPAB-2167 destroy job needs zip files to complete

andrewpearce-digital:
# Purpose

The destroy and cleanup jobs fail because the terraform evaluates the lambda zip file to generate a base64 encoded sha256sum. 

Fixes MLPAB-2167

## Approach

- pull version tag and download lambda zips to the correct path during environment cleanup job
- pull version tag and download lambda zips to the correct path during environment destroy job


hawx (approved):

acsauk (approved):
:shipit:

PR #1250, opened 2024-05-21
MLPAB-2029 - rename account level opensearch collection shared

andrewpearce-digital:
# Purpose

Avoid collisions with existing collections

Fixes MLPAB-2092

## Approach

- rename account level include "shared"


acsauk (approved):
:shipit:

PR #1253, opened 2024-05-22
MLPAB-2167 - checkout before download

andrewpearce-digital:
# Purpose

need lambda zips before cleanup

Fixes MLPAB-2167


acsauk (approved):
:shipit:

PR #1255, opened 2024-05-22
MLPAB-2127: Resequence certificate provider enter reference number for fault tolerance

acsauk:
# Purpose

Fixes MLPAB-2127

### DIFF
--- a/internal/dynamo/client.go
+++ b/internal/dynamo/client.go
@@ -432,20 +432,41 @@ func (c *Client) WriteTransaction(ctx context.Context, transaction *Transaction)
 		return errors.New("WriteTransaction requires at least one transaction")
 	}
 
-	if transaction.Errors() != nil {
-		return transaction.Errors()
-	}
-
 	var items []types.TransactWriteItem
 
-	for _, value := range transaction.Puts {
-		value.TableName = aws.String(c.table)
-		items = append(items, types.TransactWriteItem{Put: value})
+	for _, cr := range transaction.Creates {
+		values, err := attributevalue.MarshalMap(cr)
+		if err != nil {
+			return err
+		}
+
+		items = append(items, types.TransactWriteItem{Put: &types.Put{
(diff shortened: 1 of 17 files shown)


internal/dynamo/client.go
```diff
@@ -465,6 +465,21 @@ func NewTransaction() *Transaction {
...
+func (t *Transaction) Create(v interface{}) *Transaction {
+	values, err := attributevalue.MarshalMap(v)
+
+	if err != nil {
+		t.Errs = append(t.Errs, err)
+	}
+
+	t.Puts = append(t.Puts, &types.Put{
+		Item:                values,
+		ConditionExpression: aws.String("attribute_not_exists(PK) AND attribute_not_exists(SK)"),
```
acsauk, line 477: I realised we'd lost this protection during the last PR so have reinstated it here.

internal/page/certificateprovider/enter_reference_number.go
```diff
@@ -52,17 +52,13 @@ func EnterReferenceNumber(tmpl template.Template, shareCodeStore ShareCodeStore,
 					LpaID:     shareCode.LpaKey.ID(),
 				})
 
-				if _, err := certificateProviderStore.Create(ctx, shareCode.LpaOwnerKey, shareCode.ActorUID, session.Email); err != nil {
+				if _, err := certificateProviderStore.Create(ctx, shareCode, session.Email); err != nil {
```
hawx, line 55: the error handling below will need updating, which looks like it might get annoying:

```go
var ex *types.TransactionCanceledException
if !errors.As(err, &ex) {
  return err
}
for _, reason := range ex.CancellationReasons {
  if reason.Code != aws.String("ConditionalCheckFailed") {
    return err
  }
}
```

So may end up being better to wrap in `WriteTransactions` and expose a simpler error for here

hawx (approved):

PR #1261, opened 2024-05-23
MLPAB-2165 move environments over to the account level opensearch collection

andrewpearce-digital:
# Purpose

Reduce cost and build time by sharing a collection in each account

Fixes MLPAB-2165

## Approach

- use the shared opensearch collection


acsauk (approved):
:shipit:

PR #1264, opened 2024-05-24
MLPAB-2145 Only set payment completed on fee approved when paid

hawx:
Also send to lpa-store and certificate provider when paid after signing.

The issue is that we could send the `reduced-fee-requested` event without ever seeing the part payment. So I've added an `Approved` status to set the task to in that case. It will look like `InProgress` but only requires paying the missing amount.


internal/page/donor/pay.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+func Pay(
+	sessionStore SessionStore,
+	donorStore DonorStore,
+	payClient PayClient,
+	randomString func(int) string,
+	appPublicURL string,
+) Handler {
+	return func(appData page.AppData, w http.ResponseWriter, r *http.Request, donor *actor.DonorProvidedDetails) error {
```
hawx, line 21: this has always been a weird thing on its own, when really it is just another handler

acsauk: Cool - finding this should be easier now too, I always forget its in register

acsauk (approved):
:shipit:

PR #1270, opened 2024-05-29
MLPAB-2149: Resequence handleFeeApproved to support multiple events

acsauk:
# Purpose

Fixes MLPAB-2149


cmd/event-received/sirius_event_handler.go
```diff
@@ -104,24 +104,28 @@ func handleFeeApproved(ctx context.Context, client dynamodbClient, event events.
 		return err
 	}
 
+	if donor.Tasks.PayForLpa.IsCompleted() {
```
hawx, line 107:
```suggestion
	if donor.Tasks.PayForLpa.IsCompleted() || donor.Tasks.PayForLpa.IsApproved() {
```

since I've added that status

hawx (approved):

PR #1296, opened 2024-06-12
MLPAB-2069 Add versioning to hashes and exclude some fields from CheckedHash

hawx:
I think this is the best way to tackle the problem as it allows us to modify how we interpret the hashes, and in a way that each one can drift without affecting the other.

### DIFF
--- a/internal/actor/donor_provided_test.go
+++ b/internal/actor/donor_provided_test.go
@@ -1,6 +1,7 @@
 package actor
 
 import (
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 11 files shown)


internal/actor/donor_provided_test.go
```diff
@@ -20,19 +20,105 @@ var address = place.Address{
...
+				HashVersion: 0,
+				Attorneys: Attorneys{
+					Attorneys: []Attorney{
+						{DateOfBirth: date.New("2000", "1", "2")},
+					},
+				},
+			},
+			// DO NOT change these values. If a field has been added/removed you will
+			// need to handle the version gracefully by modifying HashInclude and
+			// adding another testcase.
```
hawx, line 39: of course I don't think we'll actually follow this before we start collecting real data, unless we want to practise

internal/actor/donor_provided_test.go
```diff
@@ -20,19 +21,117 @@ var address = place.Address{
...
+			},
+		}
+	}
+
+	// DO change this value to match the updates
+	const modified uint64 = 0x91100db15d995176
+
+	// DO NOT change these initial hash values. If a field has been added/removed
+	// you will need to handle the version gracefully by modifying
+	// CheckedHashInclude and adding another testcase for the new version.
```
acsauk, line 98: Should this be `CheckedHash`? Or is it referring to `toCheck.HashInclude()`?

internal/actor/donor_provided_test.go
```diff
@@ -20,19 +21,117 @@ var address = place.Address{
...
+			assert.Equal(t, donor.Hash, hash)
+			assert.False(t, donor.HashChanged())
+
+			donor.Attorneys.Attorneys[0].DateOfBirth = date.New("2001", "1", "2")
+			assert.True(t, donor.HashChanged())
+
+			err := donor.UpdateHash()
+			assert.Nil(t, err)
+			assert.Equal(t, modified, donor.Hash)
+			assert.Equal(t, uint8(1), donor.HashVersion)
```
acsauk, line 61: I wonder if we can use the test failure messages here to help new maintainers know the process for updating these tests and handling hash updates?

acsauk (commented):
Some potential to better signpost future maintainers on what they should do (outside of adding to a readme) but this seems like a sensible approach 👍

internal/actor/donor_provided_test.go
hawx: yep, the latter. I'll update

internal/actor/donor_provided_test.go
hawx: I'm hoping the comment above makes it easy enough to figure out, I don't want to be too prescriptive because I'm not certain this test structure will work forever

acsauk (approved):
:shipit:

PR #1303, opened 2024-06-19
MLPAB-2200 - enable CMK on dynamodb exports bucket

andrewpearce-digital:
# Purpose

Manage who can encrypt/decrypt objects

Fixes MLPAB-2200

## Approach

- use customer managed key for bucket encryption


acsauk (approved):
:shipit:

PR #1305, opened 2024-06-20
MLPAB-1876: Handle return codes and add vouching landing page

acsauk:
# Purpose

Fixes MLPAB-1876.


internal/page/donor/register.go
```diff
@@ -359,6 +359,11 @@ func Register(
 	handleWithDonor(page.Paths.IdentityWithOneLoginCallback, page.CanGoBack,
 		IdentityWithOneLoginCallback(commonTmpls.Get("identity_with_one_login_callback.gohtml"), oneLoginClient, sessionStore, donorStore))
 
+	handleWithDonor(page.Paths.UnableToConfirmIdentity, page.None,
+		Guidance(tmpls.Get("unable_to_confirm_identity.gohtml")))
+	handleWithDonor(page.Paths.WhatIsVouching, page.CanGoBack,
+		Guidance(tmpls.Get("unable_to_confirm_identity.gohtml")))
```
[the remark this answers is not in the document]
acsauk: No, I'd added the route in prep to assert on in the cypress test so its a placeholder for now

internal/identity/user_data.go
```diff
@@ -111,11 +111,12 @@ var charmap = map[rune][]rune{
...
-	FirstNames  string
-	LastName    string
-	DateOfBirth date.Date
-	RetrievedAt time.Time
+	OK                   bool
+	FirstNames           string
+	LastName             string
+	DateOfBirth          date.Date
+	RetrievedAt          time.Time
+	InsufficientEvidence bool
```
acsauk, line 119: I was on the fence about including this but it feels like we will need a way to differentiate between a failed ID check and someone on the vouching route.

hawx: I think it would be worth having an enum that replaces this and OK with `Passed/InsufficientEvidence/Failed` or something

internal/page/donor/identity_with_one_login_callback.go
```diff
@@ -63,6 +64,22 @@ func IdentityWithOneLoginCallback(tmpl template.Template, oneLoginClient OneLogi
 			return err
 		}
 
+		if len(userInfo.ReturnCodes) > 0 {
+			for _, c := range userInfo.ReturnCodes {
+				if c.Code == "X" {
```
hawx, line 69: I think you could clean this up by having the logic in `onelogin.Client.ParseIdentityClaim(...)` as that returns a `identity.UserData` already

hawx (approved):

PR #1310, opened 2024-06-25
MLPAB-1882 Ask for confirmation when voucher has same last name as donor

hawx:


web/template/donor/confirm_person_allowed_to_vouch.gohtml
```diff
@@ -9,27 +9,32 @@
...
-        <p class="govuk-body">
-      {{ else }}
+      {{ if .MultipleMatches }}
         <ul class="govuk-list govuk-list--bullet">
+      {{ else }}
+        <p class="govuk-body">
       {{ end }}
-      
+
+      {{ if $.MultipleMatches }}<li>{{ end }}
```
acsauk, line 18: I think the logic is a bit off here - I get an empty bullet when having an attorney and donor with the same name as the voucher:

Screenshot 2024-06-25 at 11 40 00

I'm guessing as we are filtering the `MultipleMatches` check on there being only a donor type in the slice?

hawx: ah yeah, good spot

acsauk (approved):
:shipit:

PR #1326, opened 2024-07-04
MLPAB-2166 - use new pagerduty service for production alerts

andrewpearce-digital:
# Purpose

Isolate alerts for productions resources to their own pagerduty service

Fixes MLPAB-2166

## Approach

- Use the production pagerduty service for producton account and environment


acsauk (approved):
:shipit:

PR #1327, opened 2024-07-04
MLPAB-2122: Allow donor ID mismatch updates

acsauk:
# Purpose

Fixes MLPAB-2122


internal/onelogin/user_info_test.go
```diff
@@ -270,3 +284,93 @@ func TestParseIdentityClaimWhenIdentityPublicKeyFuncError(t *testing.T) {
 	_, err := c.ParseIdentityClaim(context.Background(), UserInfo{})
 	assert.Equal(t, expectedError, err)
 }
+
+func TestCredentialAddressTransformToAddress(t *testing.T) {
```
acsauk, line 288: Felt like it was worth duplicating the tests from `place` to ensure the values assigned to `AddressDetails` don't drift at any point

web/assets/scss/main.scss
```diff
@@ -145,3 +145,7 @@ body:not(.js-enabled) .govuk-back-link {
 body.js-enabled .js-only {
     display: revert;
 }
+
+.app-08rem-font-size {
```
acsauk, line 149: Wasn't sure if it was better to be literal about this for re-use or keep it hyper-focused on where its used with something like `.app-smaller-badge-text`

andrewpearce-digital (approved):

internal/page/donor/onelogin_identity_details.go
```diff
@@ -0,0 +1,64 @@
...
+
+			f := form.ReadYesNoForm(r, "yesIfWouldLikeToUpdateDetails")
+			data.Errors = f.Validate()
+
+			if data.Errors.None() {
+				if f.YesNo.IsYes() {
+					donor.Donor.FirstNames = donor.DonorIdentityUserData.FirstNames
+					donor.Donor.LastName = donor.DonorIdentityUserData.LastName
+					donor.Donor.DateOfBirth = donor.DonorIdentityUserData.DateOfBirth
+					donor.Donor.Address = donor.DonorIdentityUserData.CurrentAddress
```
hawx, line 49: I think we should check the whole address above for a match, if we then go on to replace it all

hawx: we could simplify/remove `identity.UserData` in favour of something simpler now we do this, and just track when the identity was confirmed on `DonorProvidedDetails`

web/assets/scss/main.scss
hawx: I feel like going this small is going to be frowned upon, we should try to stick to the minimum on https://design-system.service.gov.uk/styles/type-scale/

web/template/donor/onelogin_identity_details.gohtml
```diff
@@ -0,0 +1,101 @@
...
+{{ end }}
+
+{{ define "main" }}
+  {{ $donor := .DonorProvided.Donor }}
+  {{ $userData := .DonorProvided.DonorIdentityUserData }}
+  {{ $donorFullName := (printf "%s %s" $userData.FirstNames $userData.LastName) }}
+  {{ $firstNamesMatch := eq $donor.FirstNames $userData.FirstNames }}
+  {{ $lastNameMatch := eq $donor.LastName $userData.LastName }}
+  {{ $dobMatch := eq $donor.DateOfBirth $userData.DateOfBirth }}
+  {{ $addressMatch := eq $donor.Address.Postcode $userData.CurrentAddress.Postcode }}
```
hawx, line 14: would be better to inject these from the handler, since it already does this matching logic, and then we'd only need to update one place

PR #1334, opened 2024-07-08
MLPAB-2202 - Use KMS module for kms keys

andrewpearce-digital:
# Purpose

Use module for KMS keys

Fixes MLPAB-2202

## Approach

- use module for dynamodb exports s3 bucket kms key
- use module for dynamodb kms key
- use module for opensearch kms key
- use module for reduced fees s3 bucket kms key
- use module for secrets manager kms key
- use module for sns kms key
- use module for sqs kms key
- rename all files to group them together as kms key resources


acsauk (approved):
:shipit:

PR #1336, opened 2024-07-09
MLPAB-2204 - pull av version during ur deploy job

andrewpearce-digital:
# Purpose

UR Deploy currently fails due to not receiving an input value

Fixes MLPAB-2204

## Approach

- Make sure fetch step is in deploy step's needs


acsauk (approved):
:shipit:

PR #1337, opened 2024-07-09
MLPAB-2177 - fix permission denied on decrypt in event-received

andrewpearce-digital:
# Purpose

Fix access denied on kms:decrypt with aws/lambda key

Fixes MLPAB-2177 

## Approach

- let event received role decrypt using the aws/lambda key


acsauk (approved):
:shipit:

PR #1339, opened 2024-07-09
MLPAB-2269 - add toggle for waf alb association and protect mocks

andrewpearce-digital:
# Purpose

Support turning of WAF association for ITHC

Fixes MLPAB-2269

## Approach

- Add variable to toggle WAF ALB association per environment
- protect mock pay and mock onelogin ALBs with WAF


acsauk (approved):
:shipit: 

(on the assumption the idea is we disable the WAF for ITHC)

PR #1340, opened 2024-07-09
MLPAB-2273 - use image digest from tag for mock onelogin task def

andrewpearce-digital:
# Purpose

persistent environments were not deploying new images for mock onelogin tagged latest

Fixes MLPAB-2273

## Approach

- get image digest for latest image


acsauk (approved):
:shipit:

PR #1342, opened 2024-07-10
MLPAB-2276 - don't use ci role for state access

andrewpearce-digital:
# Purpose

Apply principle of least privilege to state access

Fixes MLPAB-2276

## Approach

- use new limited state role

## Learning

- https://developer.hashicorp.com/terraform/language/settings/backends/s3


acsauk (approved):
:shipit:

PR #1347, opened 2024-07-11
MLPAB-2133:  Add  certificate provider unable to confirm ID page

acsauk:
# Purpose

I've followed the same pattern as donors here to simplify the callback logic and in prep for handling details not matching.

Fixes MLPAB-2133


internal/page/app_data.go
```diff
@@ -26,6 +26,7 @@ type AppData struct {
 	AttorneyUID       actoruid.UID
 	LoginSessionEmail string
 	SupporterData     *SupporterData
+	PublicURL         string
```
acsauk, line 29: I know we're trying to avoid stuffing too much in `AppData` but this really feels like it belongs here.

If we merge this I'll make a follow-up PR to replace function args with `AppData`.

internal/page/certificateprovider/identity_with_one_login_callback.go
```diff
@@ -79,25 +51,27 @@ func IdentityWithOneLoginCallback(tmpl template.Template, oneLoginClient OneLogi
...
-			data.DateOfBirth = userData.DateOfBirth
-			data.ConfirmedAt = userData.RetrievedAt
+		if err = certificateProviderStore.Put(r.Context(), certificateProvider); err != nil {
+			return err
+		}
 
-			if err := certificateProviderStore.Put(r.Context(), certificateProvider); err != nil {
-				return err
+		switch certificateProvider.IdentityUserData.Status {
+		case identity.StatusFailed, identity.StatusInsufficientEvidence, identity.StatusUnknown:
```
acsauk, line 60: I don't think `StatusUnknown` is possible here but it felt odd to leave it out.

internal/page/certificateprovider/unable_to_confirm_identity.go
```diff
@@ -0,0 +1,47 @@
...
+
+func UnableToConfirmIdentity(tmpl template.Template, certificateProviderStore CertificateProviderStore, lpaStoreResolvingService LpaStoreResolvingService) page.Handler {
+	return func(appData page.AppData, w http.ResponseWriter, r *http.Request) error {
+		if r.Method == http.MethodPost {
+			certificateProvider, err := certificateProviderStore.Get(r.Context())
+			if err != nil {
+				return err
+			}
+
+			certificateProvider.Tasks.ConfirmYourIdentity = actor.TaskCompleted
```
acsauk, line 27: We can see how Beth feels about this in review - setting to there's a problem or something similar didn;t feel right as the CP can still sign.

internal/page/app_data.go
hawx: we've survived this long with passing it through, so why not do that again?

hawx: `AppData` is a big bag of arguments we pass to every handler, so I think the bar for adding things should be pretty high

internal/page/certificateprovider/onelogin_identity_details.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/validation"
+)
+
+type oneloginIdentityDetailsData struct {
+	App                 page.AppData
+	Errors              validation.List
+	CertificateProvider *actor.CertificateProviderProvidedDetails
+}
+
+func OneloginIdentityDetails(tmpl template.Template, certificateProviderStore CertificateProviderStore, lpaStoreResolvingService LpaStoreResolvingService, lpaStoreClient LpaStoreClient) page.Handler {
```
hawx, line 18: I saw you did this with the donor one, could we use `OneLogin`/`one_login` instead as it looks a bit nicer and matches our previous pages. Then we can reserve the single word style for the package

internal/page/certificateprovider/onelogin_identity_details.go
```diff
@@ -0,0 +1,60 @@
...
+			}
+
+			if certificateProvider.CertificateProviderIdentityConfirmed(
+				lpa.CertificateProvider.FirstNames,
+				lpa.CertificateProvider.LastName,
+			) {
+				certificateProvider.Tasks.ConfirmYourIdentity = actor.TaskCompleted
+
+				err = certificateProviderStore.Put(r.Context(), certificateProvider)
+				if err != nil {
```
hawx, line 43: you could `if err := ...; err != nil {` this

internal/page/certificateprovider/onelogin_identity_details.go
```diff
@@ -0,0 +1,60 @@
...
+					return err
+				}
+
+				if err := lpaStoreClient.SendCertificateProviderConfirmIdentity(r.Context(), lpa.LpaUID, certificateProvider); err != nil {
```
hawx, line 47: I think this should move back, even though it makes the other handler bigger, because that is the point identity is confirmed

internal/page/certificateprovider/unable_to_confirm_identity.go
hawx: I agree with this, you've completed the work for the task even if you haven't confirmed your identity

web/template/certificateprovider/onelogin_identity_details.gohtml
```diff
@@ -0,0 +1,37 @@
+{{ template "page" . }}
```
hawx, line 1: you can delete `web/template/identity_with_one_login_callback.gohtml`

internal/page/app_data.go
acsauk: OK, that's a fair point

PR #1349, opened 2024-07-12
MLPAB-2284 Add data loss dialog for switching language

hawx:


.editorconfig
```diff
@@ -0,0 +1,12 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.{js,scss,gohtml}]
+indent_style = space
+indent_size = 2
```
hawx, line 9: I've been meaning to add this for a while, happy to go with 4 here if you prefer. I just want to stop changing it every time we edit a file the other last changed

hawx: I may try and find a better linter to make everything consistent, because the one I just tried only found the few files I've updated in this commit

andrewpearce-digital (approved):

.editorconfig
acsauk: Cool - I find it easier to parse with a 4-indent for some reason so that would be my vote

PR #1352, opened 2024-07-15
MLPAB-2114 Add legal rights page near the start

hawx:


acsauk (approved):
All good except the nav in the your details flow is missing return to task list and back link but I assumed that's because we don't have it for any other pages in that section either.

:shipit:

hawx:
Yep, that's exactly why I ignored them

PR #1358, opened 2024-07-17
MLPAB-2108 Log authentication events

hawx:


internal/telemetry/telemetry.go
```diff
@@ -16,14 +17,12 @@ import (
...
 	semconv "go.opentelemetry.io/otel/semconv/v1.12.0"
 	"go.opentelemetry.io/otel/trace"
-	"google.golang.org/grpc"
 )
 
 func Setup(ctx context.Context, resource *resource.Resource) (func(context.Context) error, error) {
 	traceExporter, err := otlptracegrpc.New(ctx,
 		otlptracegrpc.WithInsecure(),
-		otlptracegrpc.WithEndpoint("0.0.0.0:4317"),
-		otlptracegrpc.WithDialOption(grpc.WithBlock()))
```
hawx, line 26: this option is deprecated

internal/page/sign_out.go
```diff
@@ -24,6 +24,8 @@ func SignOut(logger Logger, sessionStore SessionStore, oneLoginClient OneLoginCl
 			endSessionURL = redirectURL
 		}
 
+		logger.InfoContext(r.Context(), "logout")
```
acsauk, line 27: Could it be useful to attach a session ID here to match the login event? Thinking specifically if we wanted/needed to track time active.

acsauk (commented):
Would we want to track donor [accepting access](https://github.com/ministryofjustice/opg-modernising-lpa/blob/main/internal/page/donor/enter_access_code.go) to an LPA too?

internal/page/sign_out.go
hawx: session_id will be added here from the context, it's only on login that we need to manually add it

acsauk (approved):
:shipit:

PR #1359, opened 2024-07-17
MLPAB-1574: Send email to fee payer

acsauk:
# Purpose

Fixes MLPAB-1574


internal/date/date.go
```diff
@@ -57,8 +57,8 @@ func FromTime(t time.Time) Date {
 
 	return Date{
 		year:  t.Format("2006"),
-		month: t.Format("1"),
-		day:   t.Format("2"),
```
acsauk, line 61: I've changed this to make tests a bit easier where we go from a `time.Time` to a `date.Date` as I kept getting failures for missing leading `0`s. If we can see problems with this going forward I'll find some other workaround but it felt like the least impactful.

internal/localize/localizer.go
```diff
@@ -151,6 +152,10 @@ func (l *Localizer) FormatDateTime(t time.Time) string {
 	return t.Format("2 January 2006 at 3:04pm")
 }
 
+func (l *Localizer) PenceToPounds(pence int) string {
```
hawx, line 155: this doesn't really need to be on Localizer since it is the same for both languages

internal/pay/data.go
```diff
@@ -24,7 +24,7 @@ type Link struct {
 }
 
 type CreatePaymentResponse struct {
-	CreatedDate     time.Time       `json:"created_date"`
+	CreatedDate     date.Date       `json:"created_date"`
```
hawx, line 27: image

docs suggest `time.Time` is correct here

internal/pay/data.go
```diff
@@ -68,13 +68,13 @@ type RefundSummary struct {
 }
 
 type SettlementSummary struct {
-	CaptureSubmitTime string `json:"capture_submit_time"`
-	CapturedDate      string `json:"captured_date"`
-	SettledDate       string `json:"settled_date"`
+	CaptureSubmitTime date.Date `json:"capture_submit_time"`
```
hawx, line 71: image

and this one is `time.Time`

internal/pay/data.go
```diff
@@ -68,13 +68,13 @@ type RefundSummary struct {
...
+	CapturedDate      date.Date `json:"captured_date"`
+	SettledDate       date.Date `json:"settled_date"`
 }
 
 type GetPaymentResponse struct {
-	CreatedDate time.Time `json:"created_date"`
+	CreatedDate date.Date `json:"created_date"`
```
hawx, line 77: image

and this is `time.Time` too

hawx:
image

Are/have we turned this off? (if so does this affect passing through the `Email` or getting an `Email` in the response?)

cypress/e2e/donor/payment.cy.js
```diff
@@ -1,4 +1,4 @@
-describe('Pay for LPA', () => {
+describe('Pay for LPA', {pageLoadTimeout: 8000}, () => {
```
acsauk, line 1: To account for the mock having an occasional slow response time

internal/pay/data.go
```diff
@@ -63,23 +67,33 @@ type ThreeDSecure struct {
...
 type SettlementSummary struct {
-	CaptureSubmitTime string `json:"capture_submit_time"`
-	CapturedDate      string `json:"captured_date"`
-	SettledDate       string `json:"settled_date"`
+	CaptureSubmitTime time.Time `json:"capture_submit_time"`
+	CapturedDate      date.Date `json:"captured_date"`
+	SettledDate       date.Date `json:"settled_date"`
+}
+
+type AmountPence int
```
hawx, line 80: This is a very good idea, we have been lacking a nice money type. I think renaming `Pounds()` to `String()` and `Int()` to `Pence()` would be nicer. Also if you changed `DonorProvidedDetails.FeeAmount` to return this type we'd be able to remove the `penceToPounds` template fn.

internal/pay/data.go
```diff
@@ -63,23 +67,33 @@ type ThreeDSecure struct {
...
+
+func (a AmountPence) Pounds() string {
+	return fmt.Sprintf("£%s", humanize.CommafWithDigits(float64(a)/100, 2))
```
hawx, line 83: I've started trying to avoid `fmt.Sprintf` if it can be replaced by `"x" + y`, as I think it looks less clear in these simple cases (plus premature optimisation 💪)

hawx (approved):

internal/pay/data.go
acsauk: Old habits die hard - I'd got into this in PHP ages ago but agree its generally more readable with simple strings.

internal/pay/data.go
acsauk: Nice, will update.

hawx (approved):

PR #1364, opened 2024-07-18
MLPAB-2331: Update imposter mocks to latest and use ES6

acsauk:
# Purpose

As a follow-on from #1359 I noticed imposter has included the graal VM as standard now so no need for our mock-lpa-store Dockerfile.

Fixes MLPAB-2331


docker/mock-uid/uid.js
```diff
@@ -1,19 +1,19 @@
...
 
-    for(var i = 0; i < 4; i++) {
+    for(let i = 0; i < 4; i++) {
         result.push(characters.charAt(Math.floor(Math.random() * characters.length)));
     }
 
     return result.join('');
 }
 
 //gross, but imposter doesn't support ES6 out of the box (also explains vars)
```
hawx, line 13: looks like this comment can also go

hawx (approved):

PR #1368, opened 2024-07-23
MLPAB-2198 Allow attorney to opt-out

hawx:


internal/notify/email.go
```diff
@@ -191,3 +193,15 @@ func (e PaymentConfirmationEmail) emailID(isProduction bool) string {
 
 	return "ff757818-f066-4605-8751-af481afe8a2b"
 }
+
+type AttorneyOptedOutEmail struct {
```
hawx, line 197: I assume this will have similar details to the certificate provider one when it is done, so I'm just leaving these fields in

internal/page/attorney/mock_test.go
```diff
@@ -41,3 +42,11 @@ var (
 		ActorType:   actor.TypeReplacementTrustCorporation,
 	}
 )
+
+func evalT[T any](fn func(*testing.T) T, t *testing.T) T {
```
acsauk, line 46: Cool! Now we're using funcs more in test tables this will come in handy. Could we make the name slightly more verbose for future maintainers? Something like `evalMock()` although that does signal its limited to mocks which may not be the case in every test...maybe the name is fine, I dunno 🤷

acsauk (approved):
:shipit:

internal/page/attorney/mock_test.go
hawx: I was also struggling to name this thing. If we come up with anything later it will be easy to rename

PR #1369, opened 2024-07-23
MLPAB-2225:  Enforce check and send when making LPA changes after CP has been contacted

acsauk:
# Purpose

Fixes MLPAB-2225


internal/page/donor/one_login_identity_details.go
```diff
@@ -47,6 +47,9 @@ func OneLoginIdentityDetails(tmpl template.Template, donorStore DonorStore) Hand
 					donor.Donor.LastName = donor.DonorIdentityUserData.LastName
 					donor.Donor.DateOfBirth = donor.DonorIdentityUserData.DateOfBirth
 					donor.Donor.Address = donor.DonorIdentityUserData.CurrentAddress
+					if err := donor.UpdateCheckedHash(); err != nil {
```
acsauk, line 50: Adding this for now as we probably do want to force a check and send again here but it breaks the design we built for this page. I'll raise in design review next week to get some feedback on Beth and Abbi on how we want to handle this.

hawx: :+1: that would totally make sense

hawx (approved):

PR #1372, opened 2024-07-24
Pass certificate provider into handlers

hawx:
Finally...


acsauk (approved):
Nice (boring but useful) job :shipit:

PR #1374, opened 2024-07-25
Extract some packages from actor

hawx:
This is just a proof-of-concept of how we could refactor things so `actor` stops being a dumping ground for types. I've tried to show how we can do this in stages without affecting all the code at once, but that has meant having to introduce a `temporary` package to hold the types that would cause cycles, for now.

The basic idea is to introduce a structure like:

- `{actor}/{actor}data` which holds the current `{Actor}ProvidedDetails` type (and related) (which I've renamed to `Provided`), and a `Store` which is pulled from the `app` package
- `{actor}/{actor}page` which would hold everything currently in `page/{actor}`

I've gone with the `{actor}{suffix}` naming so I don't take any good names, and I think adding 4 letters isn't that bad.

This will also force us to add more packages to hold other concepts, which is a good thing. Here I've had to move Tasks somewhere, so added a `task` package; and `ShareCodeData`, which is now in `sharecode`, this would be a good place to move the other share code related stuff.


acsauk (approved):
:shipit:

PR #1375, opened 2024-07-25
MLPAB-2343 - don't push sboms

andrewpearce-digital:
# Purpose

buildx sboms were not accessible in ECR. We'll use AWS Inspector to generate and export SBOMs for images, avoiding compatibility issues elsewhere too (lambda)

Fixes MLPAB-2343

## Approach

- don't generate and push SBOMs and provenance with buildx


acsauk (dismissed):
:shipit:

PR #1376, opened 2024-07-25
MLPAB-2213: Fix attorney and replacements back link loop and dupes

acsauk:
# Purpose

~~The issue for choose attorneys/replacement attorneys was using `history.back()` loaded the page as a POST rather than a GET so our protection of a redirect to the summary page was being skipped. This is admittedly hacky but gets the job done. If a user clicks the browser back button the bug still happens but hijacking that behaviour felt a step too far.~~

~~I initially appended a `back=1` query param to the URL when updating the widow location but I didn't need it in the end. I wonder if we want to be more intentional in what we do with users using the back link at a base handler level (at some point) and, if so, knowing when a user has clicked the link would be useful.~~

Scratch that, fixing the problem by ensuring there's always an ID present on the page.

Fixes MLPAB-2213


hawx (approved):
Much neater

PR #1379, opened 2024-07-26
Change pact test to match all statuses and include new statuses

mabely:
# Purpose

There's been a couple of changes to wording for statuses that Vega are updating in the LPA Store, to do this the pact tests need to be updated as it checks against the old wording.

[Fixes VEGA-2552](https://opgtransform.atlassian.net/browse/VEGA-2552)

## Approach

Include full list of statuses to match against


acsauk (approved):
Nice, hopefully Pact will start to feel a bit less alien once you've been through the process a couple of times 🙏 

:shipit:

PR #1403, opened 2024-08-05
Move supporter related code to packages

hawx:
In this one I've had to put the stores in `internal/supporter` to prevent an import cycle. I'm going to change the others to match as it feels neater to have the `*data` packages as just the data without any interaction.


acsauk (approved):
Agree, having the stores at the root of the packages feels more right 👍

PR #1404, opened 2024-08-05
MLPAB-2022 Use correspondent email and greeting when provided

hawx:
I will remember to update the emails when I merge this time


internal/notify/email.go
```diff
@@ -129,6 +129,7 @@ func (e DonorAccessEmail) emailID(isProduction bool) string {
 }
```
acsauk, line 129: It would be weird (but possible) for a donor to have a correspondent and be using a voucher so it may be worth adding `Greeting` to [voucher opt out](https://www.notifications.service.gov.uk/services/609565bc-0904-47a1-ae0f-9a81750193f8/templates/f21ee857-8c3e-43ee-adf2-2d9f1ff1a1a8) and [voucher refused to vouch](https://www.notifications.service.gov.uk/services/609565bc-0904-47a1-ae0f-9a81750193f8/templates/fe060b71-5c8c-46e2-9b2a-ffeee7b0e5eb) emails.

acsauk (commented):
Potentially two further templates that may need a greeting added.

internal/notify/email.go
hawx: sure, but we don't yet send those emails

acsauk: An excellent point ✅

acsauk (approved):
:shipit:

PR #1405, opened 2024-08-05
Move stores to root packages and paths for actors

hawx:
The plan for paths is to move `AppPaths` into `internal/templatefn` and change other uses to reference the constants instead of `page.Paths.Blah`


acsauk (approved):
:shipit:

PR #1412, opened 2024-08-07
MLPAB-2011 - only publish lambda when the lambda source changes

andrewpearce-digital:
# Purpose

Lambda's would previously always publish a new version even when nothing had changed about them.

Fixes MLPAB-2011

## Approach

- Use a terraform_data resource to hold a replacement trigger
- compare SHA256 to replacement trigger to determine when to publish

## Learning

- https://developer.hashicorp.com/terraform/language/resources/terraform-data

### DIFF
--- a/terraform/environment/region/modules/s3_antivirus/lambda_function.tf
+++ b/terraform/environment/region/modules/s3_antivirus/lambda_function.tf
@@ -8,6 +8,10 @@ resource "aws_lambda_layer_version" "lambda_layer" {
   provider                 = aws.region
 }
 
+resource "terraform_data" "replacement" {
+  triggers_replace = [filebase64sha256("${path.module}/myFunction.zip")]
+}
+
 resource "aws_lambda_function" "lambda_function" {
   function_name    = "zip-s3-antivirus-${data.aws_default_tags.current.tags.environment-name}"
   description      = "Function to scan S3 objects for viruses"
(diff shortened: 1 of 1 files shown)


terraform/environment/region/modules/s3_antivirus/lambda_function.tf
```diff
@@ -18,7 +22,7 @@ resource "aws_lambda_function" "lambda_function" {
   runtime          = "provided.al2023"
   timeout          = 300
   memory_size      = 4096
-  publish          = true
+  publish          = filebase64sha256("${path.module}/myFunction.zip") != terraform_data.replacement.triggers_replace[0] ? true : false
```
hawx, line 25: wouldn't `!=` give you a `true` or `false`?

hawx (dismissed):

terraform/environment/region/modules/s3_antivirus/lambda_function.tf
andrewpearce-digital: If there's a difference between the current sha and what's stored in the trigger, publish = true, is what' I'm going for.

andrewpearce-digital: oh I see what you mean, yes, I could leave it as `a != b`

acsauk (approved):
:shipit:

PR #1414, opened 2024-08-08
Move more things

hawx:
- the lpastore data types to `lpadata`
- the progress tracker to `task`
- makes `page.MultiReadCloser` private

my current goal is to remove the non-handler/path stuff from `page`


acsauk (approved):
Looking good :shipit:

PR #1434, opened 2024-08-20
MLPAB-1909 Add pages for voucher identity check

hawx:


internal/page/fixtures/voucher.go
```diff
@@ -32,6 +33,7 @@ func Voucher(
 	progressValues := []string{
 		"confirmYourName",
 		"verifyDonorDetails",
+		"confirmYourIdentity",
```
acsauk, line 36: Do we want to add this to the fixtures page too?

acsauk (approved):
:shipit:

internal/page/fixtures/voucher.go
hawx: I did and I forgot, I'll add it in the next one

PR #1453, opened 2024-08-29
MLPAB-2355 Expire identity check after 6 months

hawx:
The idea for the design is to chunk the work by day, putting in dynamo using the day for the PK. The runner will run every 6 hours (current arbitrary choice) and work through the items for that day. It changes the keys of items that have been picked up to have the suffix `#HANDLED`.

Things it should probably do in the future:
- be in a lambda, that can then be triggered with eventbridge on schedule
- remove successfully run items, or report on the `#HANDLED` item what the outcome was (so we can figure out what worked/didn't, and re-run failures)
- maybe deal with the possibility that it won't finish a days work in a day? (I doubt this would ever happen, and if it did we should be able to change the affected PKs to a future day to get them picked up)
- x-ray tracing


internal/scheduled/runner.go
```diff
@@ -0,0 +1,170 @@
...
+			continue
+		}
+	}
+}
+
+func (r *Runner) step(ctx context.Context) error {
+	r.waiter.Reset()
+
+	for {
+		row, err := r.store.Pop(ctx, r.now())
```
hawx, line 101: to test locally you can either change the PK of the item in dynamo, or do `r.now().AddDate(0,6,0)` here

internal/scheduled/event.go
```diff
@@ -0,0 +1,27 @@
...
+// A Event specifies an action to take in the future.
+type Event struct {
+	PK dynamo.ScheduledDayKeyType
+	SK dynamo.ScheduledKeyType
+	// CreatedAt is when the Row was created
+	CreatedAt time.Time
+	// LockedUntil is set to show the Row is running, it cannot be retrieved until this time has passed
+	LockedUntil time.Time
+	// HandledAt is set when the Row has finished running
+	HandledAt time.Time
```
acsauk, line 18: Are these being used yet?

acsauk (commented):
Nice approach - potentially a couple of props that haven't been used but otherwise good to go.

internal/scheduled/event.go
hawx: ah nope, these are leftover from a previous way I was going to do things

acsauk (approved):
:shipit:

PR #1454, opened 2024-08-29
Update mocks to latest mockery

hawx:
There is an issue currently with following aliases, so this was generated with `GODEBUG=gotypesalias=0 mockery` as per
https://github.com/vektra/mockery/issues/803#issuecomment-2287198024


acsauk (approved):
:shipit: missed this before pushing my update branch!

PR #1461, opened 2024-09-03
MLPAB-2134: Track failed vouches and vary guidance based on attempts

acsauk:
# Purpose

Fixes MLPAB-2134


internal/voucher/voucherpage/identity_with_one_login_callback.go
```diff
@@ -56,6 +56,17 @@ func IdentityWithOneLoginCallback(oneLoginClient OneLoginClient, sessionStore Se
...
 
 		if !provided.IdentityConfirmed() {
+			donor, err := donorStore.GetAny(r.Context())
+			if err != nil {
+				return err
+			}
+
+			donor.FailedVouchAttempts++
+
+			if err := donorStore.Put(r.Context(), donor); err != nil {
```
hawx, line 66: this could end up incrementing again if the `SendActorEmail` fails and the person refreshes the page, so maybe move this down

hawx:
We also really need to do the sending voucher code ticket, would be nice to verify this end-to-end

hawx (approved):

PR #1464, opened 2024-09-04
MLPAB-2407 Use published onelogin public key for verifying identity jwt

hawx:


andrewpearce-digital (approved):

internal/onelogin/did_test.go
```diff
@@ -0,0 +1,369 @@
...
+			Body: io.NopCloser(strings.NewReader(validDIDDocument)),
+		}, nil).
+		Once()
+	doer.EXPECT().
+		Do(mock.Anything).
+		Return(&http.Response{
+			StatusCode: http.StatusOK,
+			Body:       io.NopCloser(strings.NewReader(validDIDDocument)),
+		}, nil).
+		Run(func(req *http.Request) {
```
acsauk, line 68: Ah ha - I think this is the step I was missing...so by calling `Run()` you can sequence some actions to happen after this mock does what it needs in the test?

PR #1470, opened 2024-09-06
MLPAB-1897: Send vouching access code to donor

acsauk:
# Purpose

Fixes MLPAB-1897


package.json
```diff
@@ -35,5 +35,6 @@
     },
     "engines": {
         "node": "20.17.0"
-    }
+    },
+    "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
acsauk, line 39: This was required to get yarn to respect the version of node I had installed after running:

`bash
corepack enable
asdf reshim nodejs
`

I'm not 100% sure of side-effects of defining the version we're using here although it does look like we'll need to define this if we ever want to upgrade yarn above v1.

andrewpearce-digital (approved):

PR #1487, opened 2024-09-12
MLPAB-2405 Update event tests

hawx:
I've pulled the latest versions and added the missing ones. I've also removed sending `previous-application-linked` as it was never defined, or planned to be used in Sirius.

I also added a test to check every schema that has been downloaded has been tested, and that we have a schema for every event we send. I think checking that everything is up-to-date would be annoying, so that is the only missing thing we have to remember.


andrewpearce-digital (approved):

internal/event/client.go
```diff
@@ -14,6 +15,18 @@ import (
 
 const source = "opg.poas.makeregister"
 
+var events = map[any]string{
+	(*UidRequested)(nil):               "uid-requested",
```
acsauk, line 19: This is cool, I've wanted to have a map of types in the past but could never work out how to do it 🙌

acsauk (approved):
:shipit:

PR #1505, opened 2024-09-23
MLPAB-2470 Add watermark to printed contextual lpa page

hawx:


web/template/layout/contextual-lpa.gohtml
```diff
@@ -1,6 +1,6 @@
-<!-- TODO: remove no_actions from summary-row as it adds additional spacing not needed -->
```
hawx, line 1: I did this TODO already, but forgot I'd left myself a note...

PR #1518, opened 2024-09-27
MLPAB-2504 Make lowerFirst and possessive available in translation strings

hawx:


acsauk (approved):
I checked through the translation file for any missed uses of `lowerFirst` and `possessive` and came across some unused translations:

```
myLpa
theirLpa
```

Unless we're building the string in code somewhere we should be able to drop these from the translation files.

PR #1521, opened 2024-09-30
MLPAB-2287: Fix ID and sign task list and registering with COP routing

acsauk:
# Purpose

Fixes MLPAB-2287


internal/donor/donorpage/task_list.go
```diff
@@ -192,6 +192,15 @@ func taskListSignSection(provided *donordata.Provided) taskListSection {
 	case identity.StatusFailed:
 		signPath = donor.PathRegisterWithCourtOfProtection
 
+		if provided.RegisteringWithCourtOfProtection {
+			signPath = donor.PathReadYourLpa
+			if !provided.SignedAt.IsZero() && provided.WitnessedByCertificateProviderAt.IsZero() {
```
hawx, line 197: you could rotate these, so `if !Witnessed.IsZero() { YouHaveSubmitted } else if !SignedAt.IsZero { Witnessing }`

acsauk: Totally wrong - I'll update and switch the ifs for clarity.

I raised the increasing complexity of the logic for this item in the task list in collab time today and we agreed once the next batch of tickets in flight are completed we'd play back what we have to confirm it's all as expected.

hawx (approved):

PR #1530, opened 2024-10-03
MLPAB-2170 Send notification when voucher confirms

hawx:


cypress/e2e/voucher/confirm-your-name.cy.js
```diff
@@ -1,6 +1,6 @@
 describe('Confirm your name', () => {
     beforeEach(() => {
-        cy.visit('/fixtures/voucher?redirect=/confirm-your-name');
+        cy.visit('/fixtures/voucher?redirect=/confirm-your-name&progress=skip');
```
acsauk, line 3: Was this supposed to be `progress=` to match up with the changes in voucher fixtures?

internal/lpastore/lpa.go
```diff
@@ -495,7 +495,7 @@ func FromDonorProvidedDetails(l *donordata.Provided) *lpadata.Lpa {
 		Restrictions:                             l.Restrictions,
 		WhenCanTheLpaBeUsed:                      l.WhenCanTheLpaBeUsed,
 		LifeSustainingTreatmentOption:            l.LifeSustainingTreatmentOption,
-		SignedAt:                                 l.SignedAt,
+		SignedAt:                                 l.WitnessedByCertificateProviderAt,
```
acsauk, line 498: I understand this change but it bothers me that there's some potential for a drift between signing and witnessing dates here if a CP fails to witness after the donor signs that could be confusing. To keep in sync, do we need to add `WitnessedByCertificateProviderAt` to the Lpa store data? (not as part of this ticket)

acsauk (commented):
One potential change on a fixture link

cypress/e2e/voucher/confirm-your-name.cy.js
hawx: nope, purposefully using one that doesn't exist so it doesn't do the sharecode sending

internal/lpastore/lpa.go
hawx: potentially, but we'll never drive logic off our `SignedAt` date as it isn't _really_ signed at that point

cypress/e2e/voucher/confirm-your-name.cy.js
hawx: I was using one that didn't exist to skip the behaviour of it being unset, but I've rewritten how it works to maybe be more obvious

acsauk (approved):
:shipit:

PR #1535, opened 2024-10-08
MLPAB-2471 Send witnessed times to lpa-store and show in contextual lpa

hawx:


internal/lpastore/lpadata/lpa.go
```diff
@@ -69,6 +81,12 @@ type Lpa struct {
 	Voucher Voucher
 }
 
+// SignedForDonor returns true if the Lpa has been signed and witnessed for the donor.
+func (l *Lpa) SignedForDonor() bool {
```
acsauk, line 85: I am now worry free 👍

hawx: 🙌

PR #1539, opened 2024-10-08
MLPAB-2566 - set dead letter queue for all custom bus eventbridge targets

andrewpearce-digital:
# Purpose

Capture failed event delivery using a SQS queue on the custom event bus

Fixes MLPAB-2566

## Approach

- output SQS DLQ on the custom bus
- Use the DLQ for event-received event bus targets
- add a policy to the SQS queue to allow all rules/targets to deliver to queue

## Learning

- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudwatch_event_target#dead_letter_config
- https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-rule-dlq.html
- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/sqs_queue_policy


acsauk (approved):
:shipit:

PR #1540, opened 2024-10-08
fix dep reference

andrewpearce-digital:
# Purpose

incorrect reference to the package

Fixes MLPAB-2557

## Approach

- use module source as reference
- fix renovate config warning about `prPriority` in config `major`


acsauk (approved):
:shipit:

PR #1541, opened 2024-10-08
MLPAB-2408: Add scheduled task schedule runner

acsauk:
# Purpose

Fixes MLPAB-2408


internal/scheduled/runner.go
```diff
@@ -73,45 +73,49 @@ func NewRunner(logger Logger, store ScheduledStore, donorStore DonorStore, notif
...
 
 	for {
 		row, err := r.store.Pop(ctx, r.now())
 		if errors.Is(err, dynamo.NotFoundError{}) {
+			r.logger.InfoContext(ctx, "not found")
+
 			return nil
 		} else if errors.Is(err, dynamo.ConditionalCheckFailedError{}) {
 			r.logger.InfoContext(ctx, "runner conditional check failed")
-			if err := r.waiter.Wait(); err != nil {
```
hawx, line 106: you'll want some kind of waiting, otherwise it may just keep hammering dynamo when there is an underlying problem

terraform/environment/region/modules/schedule_runner/lambda.tf
```diff
@@ -0,0 +1,197 @@
...
+    sid       = "allowApiAccess"
+    effect    = "Allow"
+    resources = var.allowed_api_arns
+    actions = [
+      "execute-api:Invoke",
+    ]
+  }
+}
+
+resource "aws_scheduler_schedule" "schedule_runner_hourly" {
```
andrewpearce-digital, line 41: I think you're missing a region provider here, it'll pass and then fail to apply because it doesn't have permission to create that resource in the identity account (your user that is used when no provider is given)

terraform/environment/region/modules/schedule_runner/lambda.tf
```diff
@@ -0,0 +1,162 @@
...
+    SEARCH_ENDPOINT            = var.search_endpoint
+    SEARCH_INDEX_NAME          = var.search_index_name
+    SEARCH_INDEXING_DISABLED   = 1
+  }
+  image_uri            = "${var.lambda_function_image_ecr_url}:${var.lambda_function_image_tag}"
+  aws_iam_role         = var.schedule_runner_lambda_role
+  environment          = data.aws_default_tags.current.tags.environment-name
+  kms_key              = data.aws_kms_alias.cloudwatch_application_logs_encryption.target_key_arn
+  iam_policy_documents = [data.aws_iam_policy_document.schedule_runner.json]
+  timeout              = 360
```
andrewpearce-digital, line 19: timeout can be up to 900 seconds (15 minutes)

terraform/environment/region/modules/schedule_runner/lambda.tf
```diff
@@ -0,0 +1,162 @@
...
+      "kms:Decrypt",
+      "kms:GenerateDataKey",
+      "kms:GenerateDataKeyPair",
+      "kms:GenerateDataKeyPairWithoutPlaintext",
+      "kms:GenerateDataKeyWithoutPlaintext",
+      "kms:DescribeKey",
+    ]
+  }
+
+  #   statement {
```
andrewpearce-digital, line 112: this can go if the deployed env works works

.github/workflows/docker_job.yml
```diff
@@ -53,6 +53,10 @@ jobs:
             path: ./docker/mock-pay/Dockerfile
             trivyignores: ./docker/mock-pay/.trivyignore.yaml
             platforms: linux/amd64
+          - ecr_repository: modernising-lpa/schedule-runner
+            name: schedule-runner
+            path: ./docker/schedule-runner/Dockerfile
```
andrewpearce-digital, line 58: you'll need to reference the ignore file here

docker/schedule-runner/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+ENTRYPOINT ["./schedule-runner"]
+
+FROM public.ecr.aws/lambda/provided:al2023.2024.10.14.12 AS production
+
+WORKDIR /app
+COPY --link  docker/install_lambda_insights.sh /app/
+
+RUN chmod +x "/app/install_lambda_insights.sh" \
+  && /app/install_lambda_insights.sh ${TARGETPLATFORM}
```
andrewpearce-digital, line 29: this is asking you to do this...
```suggestion
  && /app/install_lambda_insights.sh "${TARGETPLATFORM}"
```

terraform/environment/region/modules/schedule_runner/lambda.tf
```diff
@@ -0,0 +1,150 @@
...
+    security_group_ids = var.vpc_config.security_group_ids
+  }
+  providers = {
+    aws.region = aws.region
+  }
+}
+
+resource "aws_scheduler_schedule" "schedule_runner_twenty_minutes" {
+  name                = "schedule-runner-hourly"
+  schedule_expression = "rate(20 minute)"
```
hawx, line 32: `-hourly` `twenty_minutes`, probably don't need as often right now, or make it configurable per env?

cmd/schedule-runner/main.go
```diff
@@ -0,0 +1,92 @@
...
+		eventBusName          = cmp.Or(os.Getenv("EVENT_BUS_NAME"), "default")
+		notifyBaseURL         = os.Getenv("GOVUK_NOTIFY_BASE_URL")
+		notifyIsProduction    = os.Getenv("GOVUK_NOTIFY_IS_PRODUCTION") == "1"
+		searchEndpoint        = os.Getenv("SEARCH_ENDPOINT")
+		searchIndexName       = cmp.Or(os.Getenv("SEARCH_INDEX_NAME"), "lpas")
+		searchIndexingEnabled = os.Getenv("SEARCH_INDEXING_DISABLED") != "1"
+		tableName             = os.Getenv("LPAS_TABLE")
+	)
+
+	logger := slog.New(slog.NewJSONHandler(os.Stdout, nil).
```
hawx, line 35: it would be good to have x-ray hooked up too, but could make a new ticket for that

internal/app/app.go
```diff
@@ -53,6 +53,7 @@ type DynamoClient interface {
 	AllByPartialSK(ctx context.Context, pk dynamo.PK, partialSK dynamo.SK, v interface{}) error
 	AllBySK(ctx context.Context, sk dynamo.SK, v interface{}) error
 	AllKeysByPK(ctx context.Context, pk dynamo.PK) ([]dynamo.Keys, error)
+	AnyByPK(ctx context.Context, pk dynamo.PK, v interface{}) error
```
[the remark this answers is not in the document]
acsauk: I dunno...if we're expecting only a single result back I'd rather not blindly take the first result without knowing something has gone wrong. With the best will in the world we'll probably write some bad data at some point. By keeping this stricter we'll know about it sooner.

docker/schedule-runner/Dockerfile
acsauk: Haha got it - as I added quotes around the insights script it did seem a bit unlikely to have globbing on a regular string!

terraform/environment/region/modules/schedule_runner/lambda.tf
acsauk: I'll revert to hourly. We can trigger a manual invoke on an env if we need to test something out.

cmd/schedule-runner/main.go
acsauk: Good call - I'll add a ticket to the backlog.

hawx (approved):

PR #1543, opened 2024-10-09
MLPAB-2569 Disable mock pay on demo env

hawx:
All the environments on development use mock-pay, so I'm thinking it will be fine to just change the existing secret to a new API key for the correct account and only demo will use that. Unless anyone can think of a reason we'd want to use another test account?


acsauk (approved):
Makes sense, cant think of a use case for an additional test account :shipit:

PR #1552, opened 2024-10-15
MLPAB-2599 - Apply a security standard to Dockerfiles

andrewpearce-digital:
# Purpose

Apply a security standard to dockerfiles

Fixes MLPAB-2599

## Approach
Trivy scan returned findings for 
- HIGH: Specify at least 1 USER command in Dockerfile with non-root user as argument
- LOW: Add HEALTHCHECK instruction in your Dockerfile

For images used in lambda functions
> Note that the example Dockerfile does not include a [USER instruction](https://docs.docker.com/reference/dockerfile/#user). When you deploy a container image to Lambda, Lambda automatically defines a default Linux user with least-privileged permissions. This is different from standard Docker behavior which defaults to the root user when no USER instruction is provided.

Note: This PR add a Trivy check in the docker build job for security related configuration issues in Dockerfiles that are deployed to AWS (ECS and Lambda). Failing the check fails the workflow.
___ 

#### event-received
- use al2023 instead of al2
- ignore AVD-DS-0002 - Lambda creates a docker USER with least-privilege permissions.

#### create_s3_replication_job
- ignore AVD-DS-0002 - Lambda creates a docker USER with least-privilege permissions.

#### mock-pay
- Set non-root user

## Learning

- https://docs.aws.amazon.com/linux/al2/ug/what-is-amazon-linux.html
- https://aquasecurity.github.io/trivy/dev/docs/configuration/filtering/
- https://github.com/aquasecurity/trivy-action
- https://github.com/outofcoffee/imposter/blob/main/distro/base/Dockerfile


acsauk (approved):
:shipit:

PR #1555, opened 2024-10-16
MLPAB-2606 - Apply standards from CIS Docker Baseline 1.7.0 to ECS images

andrewpearce-digital:
# Purpose

Apply remaining standards to ECS images

Fixes MLPAB-2606

## Approach

- CIS 4.5 - enable docker content trust on build and push jobs for images pulled during build
- mlpa app image
  - Use ual container hardening script for alpine 
  - CIS 4.8 remove setuid and setgid permissions
  - CIS 4.3 Ensure that unnecessary packages are not installed in the container
  - CIS 5.8 Ensure privileged ports are not mapped within containers
  - CIS 5.9 Ensure that only needed ports are open on the container
- mock-pay container 
  - CIS 5.8 Ensure privileged ports are not mapped within containers
  - CIS 5.9 Ensure that only needed ports are open on the container 

## Learning

- https://docs.docker.com/engine/security/trust/trust_automation/
- https://github.com/ministryofjustice/opg-use-an-lpa/blob/main/scripts/docker_hardening/harden.sh


.github/workflows/workflow_pr.yml
```diff
@@ -135,10 +138,14 @@ jobs:
 
   ui_tests_pr_env:
     name: Run Cypress UI Tests On PR Environment
-    if: always() &&
-      needs.pr_deploy.result == 'success'
+    if: always()
+    # if: always() &&
+    #   needs.pr_deploy.result == 'success'
```
acsauk, line 143: Do we need this still?

acsauk (commented):
Potentially some commented out code we could drop

.github/workflows/workflow_pr.yml
andrewpearce-digital: thank you, yes we need that

acsauk (approved):
:shipit:

docker/mlpa/Dockerfile
```diff
@@ -54,9 +54,17 @@ COPY --from=build-env /go/bin/mlpab mlpab
...
+
+RUN addgroup -S user && \
+  adduser -S -g user user && \
+  chown -R user:user mlpab web/template web/static web/robots.txt
+
+COPY scripts/docker_hardening/alpine_image_hardening.sh /harden.sh
+
+RUN /harden.sh && rm /harden.sh
+
+USER user
```
mattmachell, line 66: Might want to make this user more specifically named, like `mlpaapp` or similar?

mattmachell: Ah, I see. Maybe `appuser` then. Naming to `user` just feels a bit unclear on intent to me.

andrewpearce-digital: I need the name of the user to be agnostic of the container it's used in. The intent of the user is seen a couple of lines above where its given ownership of the the container's app code

PR #1556, opened 2024-10-16
MLPAB-2439 Prevent donor changes to a signed LPA

hawx:


internal/donor/store.go
```diff
@@ -323,6 +323,11 @@ func (s *Store) Put(ctx context.Context, donor *donordata.Provided) error {
 		return nil
 	}
 
+	appData := appcontext.DataFromContext(ctx)
+	if appData.IsDonor() && !donor.CanChange() && donor.CheckedHashChanged() {
+		return errors.New("donor tried to change a signed LPA")
```
hawx, line 328: I've raised https://opgtransform.atlassian.net/browse/MLPAB-2610 to look at what we actually want to do from a design perspective, as this isn't obvious to a donor

acsauk (approved):
:shipit:

PR #1557, opened 2024-10-16
MLPAB-2592: Don't request UID if already exists

acsauk:
# Purpose

The voucher part of the reproducible steps was red herring, it's the existing bug we already know about where a UID is requested every time the lpa-type form is submitted forcing a re-check of data. I was considering allow listing UID as its not something a donor can change but this achieves the same thing.
 
Fixes MLPAB-2592


hawx (commented):
ha, didn't realise this was happening. Could you also put a guard at the start of the `uid-requested` handler, so if the event is handled twice it doesn't overwrite the first?

hawx (dismissed):

hawx (approved):

PR #1558, opened 2024-10-17
Namespace scheduler name to env and region

acsauk:
# Purpose

To fix the pipeline.

### DIFF
--- a/terraform/environment/region/modules/schedule_runner/lambda.tf
+++ b/terraform/environment/region/modules/schedule_runner/lambda.tf
@@ -28,7 +28,7 @@ module "schedule_runner" {
 }
 
 resource "aws_scheduler_schedule" "schedule_runner_hourly" {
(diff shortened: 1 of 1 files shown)


terraform/environment/region/modules/schedule_runner/lambda.tf
```diff
@@ -28,7 +28,7 @@ module "schedule_runner" {
 }
 
 resource "aws_scheduler_schedule" "schedule_runner_hourly" {
-  name                = "schedule-runner-hourly"
+  name                = "schedule-runner-hourly-${data.aws_default_tags.current.tags.environment-name}-${data.aws_region.current.name}"
```
andrewpearce-digital, line 31: I don't think we need region as part of the name 

```suggestion
  name                = "schedule-runner-hourly-${data.aws_default_tags.current.tags.environment-name}"
```

andrewpearce-digital (approved):

PR #1560, opened 2024-10-17
MLPAB-2290 Update content of what happens next cop page

hawx:


lang/cy.json
```diff
@@ -297,17 +297,28 @@
...
     },
+    "opgWillContactAttorneys": {
+        "one": "Welsh {{.FullNames}}",
+        "two": "Welsh {{.FullNames}}",
+        "few": "Welsh {{.FullNames}}",
+        "many": "Welsh {{.FullNames}}",
+        "other": "Welsh {{.FullNames}}"
+    },
     "onceEveryoneHasSigned": "Unwaith y bydd pawb wedi llofnodi a phan fydd eich LPA wedi’n cyrraedd, byddwn yn hysbysu unrhyw bobl y byddwch yn dewis rhoi gwybod iddynt. Bydd cyfnod aros o 4 wythnos yn dechrau. Mae hyn yn ofynnol yn ôl y gyfraith, er mwyn caniatáu i unrhyw un sydd â gwrthwynebiad i’w codi.",
+    "onceEveryoneHasSignedOpgWill": "Welsh",
```
acsauk, line 316: Should this (and other mid-key instances) be capitalised for consistency?

acsauk (commented):
I may have missed some comments in the figma but bullets 3 and 4 in the ~unsigned~ signed version looks slightly different than whats in the design:

Screenshot 2024-10-18 at 10 56 33

lang/cy.json
hawx: I don't think we've been consistent with the keys anyway, the only one I spotted allcaps was `theOPG`, which isn't even used...

acsauk (approved):
:shipit:

PR #1572, opened 2024-10-24
MLPAB-2620 - fix issue with access to manifests bucket

andrewpearce-digital:
# Purpose

Resolve unauthorised api call errors for list bucket actions when using the batch manifests role

Fixes MLPAB-2620

## Approach

- Add list bucket action for batch manifests

## Learning

- https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazons3.html


acsauk (approved):
:shipit:

PR #1573, opened 2024-10-24
MLPAB-2624 - use pr number only for terraform workspace names

andrewpearce-digital:
# Purpose

Environment naming is longer than is useful. Generally we search for environment resources using the PR number, and the additional text is being trimmed at `mlpabxxx`, which is less useful for spotting your own env resources.

Fixes MLPAB-2624

## Approach

- in workspace generator, only use PR number unless it's for a weblate PR, then use the workspace name weblate


acsauk (approved):
:shipit:

PR #1574, opened 2024-10-25
MLPAB-2456 Show certificate provider and attorney the contextual LPA

hawx:
This also has a change to xray that I've been playing with. Instead of having a segment with the event name

image

put the information on the existing Events span

image

I really want to do something similar for Notify, but can't figure out a nice way to get it to work.


acsauk (approved):
Nice - the event feels like a better structure using the span.

PR #1582, opened 2024-10-29
MLPAB-2607: Add xray to schedule runner lambda

acsauk:
# Purpose

We've (well, @andrewpearce-digital) grabbed the otel collector lambda layer we're using here from an ECR image and we've decided to persist it in the repo for now. When we come to apply this to other lambdas we may want to consider automating this step to ensure we get updates as it could quickly get out date.

I've added [MLPAB-2631](https://opgtransform.atlassian.net/browse/MLPAB-2631) to the backlog to look at how we catch otel errors and gracefully end the lambda (I tried and failed to find the right way to hook into the process). 

Fixes MLPAB-2607

[MLPAB-2631]: https://opgtransform.atlassian.net/browse/MLPAB-2631?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/telemetry/telemetry.go
```diff
@@ -6,40 +6,83 @@ import (
...
 	"go.opentelemetry.io/otel/trace"
 )
 
-func Setup(ctx context.Context, resource *resource.Resource) (func(context.Context) error, error) {
-	traceExporter, err := otlptracegrpc.New(ctx,
-		otlptracegrpc.WithInsecure(),
-		otlptracegrpc.WithEndpoint("0.0.0.0:4317"))
-	if err != nil {
-		return nil, fmt.Errorf("failed to create new OTLP trace exporter: %w", err)
+func Setup(ctx context.Context, resource *resource.Resource, stdOutOverride bool) (func(context.Context) error, error) {
```
hawx, line 26: could remove the `stdOutOverride` now

hawx (commented):
it would also be good to use `telemetry.NewSlogHandler` with the logger so you get logs associated with the trace

hawx (approved):
good call on adding the runbook too

PR #1584, opened 2024-10-30
Add removed translation key back in

hawx:
I think this is what is tripping weblate up, as the key has been edited in a weblate commit but removed by one of my PRs. So hopefully adding it back will unblock weblate, and then I'll be able to delete again once that commit is merged.


acsauk (approved):
:shipit:

PR #1586, opened 2024-10-31
MLPAB-2609 Remove continue when return to task list is only option

hawx:
Also fix an issue with the task list going to the wrong page when going to COP and signing the LPA.


acsauk (approved):
:shipit:

PR #1588, opened 2024-10-31
MLPAB-2608 Prevent writes of items where only a single instance of the SK should…

hawx:
I've taken the approach of using transactions to write records that will then be checked for uniqueness. 

For example, to prevent having muliple donors on an LPA, we now write:

|PK|SK|
|---|---|
|LPA#uuid|DONOR#sub|
|LPA#uuid|RESERVED#DONOR#|

This means we can't put another `DONOR#other-sub` on the same LPA, as the `(LPA#uuid, RESERVED#DONOR#)` record would conflict. I'm using a `RESERVED#` prefix so that these records aren't returned in our existing queries. This same technique is used to ensure only one certificate provider per LPA, and I'm writing `(UID#uid, METADATA#)` to prevent reusing UIDs.

To ensure sharecodes are unique, I've added a `dynamo.Client.CreateOnly` method that will check the `PK` does not exist (as opposed to the `(PK, SK)` pair in `Create`).


cmd/event-received/sirius_event_handler.go
```diff
@@ -241,6 +222,23 @@ func handleDonorSubmissionCompleted(ctx context.Context, client dynamodbClient,
 		return fmt.Errorf("failed to send share code to certificate provider: %w", err)
 	}
 
+	lpaID := uuidString()
+
+	transaction := dynamo.NewTransaction().
```
hawx, line 227: reordered here as the other parts are retryable

acsauk (commented):
Smart solution - my only query is should we apply this to the Voucher record as well?

PR #1617, opened 2024-11-13
MLPAB-2462 Add role=img to printer icon

hawx:
I don't have an iOS device to try this on, but it did stop it announcing as a group on Mac Safari so maybe it will work?

### DIFF
--- a/internal/sesh/dynamo_store.go
+++ b/internal/sesh/dynamo_store.go
@@ -2,6 +2,7 @@ package sesh
 
 import (
 	"context"
+	"encoding/base64"
 	"net/http"
 
 	"github.com/gorilla/securecookie"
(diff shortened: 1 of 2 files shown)


internal/sesh/dynamo_store.go
```diff
@@ -67,7 +68,7 @@ func (s *DynamoStore) Save(r *http.Request, w http.ResponseWriter, session *sess
 	}
 
 	if session.ID == "" {
-		session.ID = string(securecookie.GenerateRandomKey(32))
+		session.ID = base64.RawStdEncoding.EncodeToString(securecookie.GenerateRandomKey(32))
```
hawx, line 71: I just noticed this when looking in DynamoDB 🤦

andrewpearce-digital (approved):

PR #1619, opened 2024-11-13
MLPAB-2278: Update payment confirmation content

acsauk:
# Purpose

Fixes MLPAB-2278


docs/runbooks/managing_node_versions.md
```diff
@@ -17,3 +17,9 @@ From here you can install Node dependencies with Yarn.
 ```shell
 brew install yarn
 ```
+
+If you encounter unrecognised versions of node after adding them to asdf you may need reinstall yarn for node re-shims to be recognised:
+
+```shell
+npm install -g yarn
```
hawx, line 24: is there any point to using yarn these days / would it simplify things if we just used npm?

hawx (approved):

docs/runbooks/managing_node_versions.md
acsauk: Fair point - I'd like to look at speed differences of the most recent version of yarn (we're on 1.22, looks like v4 came out last year), npm and pnpm to see if we can shave some build time down but I'm not overly invested in any specific option.

acsauk: https://opgtransform.atlassian.net/browse/MLPAB-2685

PR #1629, opened 2024-11-18
MLPAB-2454 Collect identity info when they have completed at post office

hawx:


acsauk (approved):
All looks good but I wasn't able to find a journey through to see the CP ID pending page when they have provided their certificate as the LPA drops off the dashboard. Has this been covered on another ticket?

hawx:
I raised https://opgtransform.atlassian.net/browse/MLPAB-2689 for it, for now you can check it works by changing the URL at the end of the journey

PR #1630, opened 2024-11-18
MLPAB-2605: Add runner performance metrics

acsauk:
# Purpose

To be followed by MLPAB-2609 to add a test to the pipeline with a large amount of scheduled tasks and make infra changes to allow metrics to be collected.

Fixes MLPAB-2605 

## Approach

I started with some basic logs with processing time but it was a pain to get the data out so I dumped into a few high-level metrics. Very open to feedback on what/how we're measuring here as I've not done much in this space before.


cmd/scheduled-task-adder/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+var taskCount int
+var entryPoint string
+
+type TaskCountEvent struct {
+	TaskCount int `json:"taskCount"`
+}
+
+func init() {
+	flag.IntVar(&taskCount, "taskCount", 0, "Number of scheduled tasks to generate")
```
acsauk, line 31: I'd originally made this as a local script to run so have kept the option of building and running it outside of docker. I'm not sure how much use it will get in this form so if we think YAGNI I can simplify to be just a lambda.

docker/localstack/localstack-init.sh
```diff
@@ -26,6 +26,14 @@ awslocal dynamodb create-table \
          --provisioned-throughput ReadCapacityUnits=1000,WriteCapacityUnits=1000 \
          --global-secondary-indexes file://dynamodb-lpa-gsi-schema.json
 
+awslocal dynamodb create-table \
```
acsauk, line 29: For when we run a pipeline performance test on a large amount of tasks.

cmd/schedule-runner/main.go
```diff
@@ -81,11 +86,27 @@ func handleRunSchedule(ctx context.Context) error {
...
+	if metricsEnabled {
+		if Tag == "" {
+			Tag = os.Getenv("TAG")
+		}
+
+		client := cloudwatch.NewFromConfig(cfg)
+
+		metricsClient := telemetry.NewMetricsClient(client, Tag)
+
+		if err = metricsClient.PutMetrics(ctx, &metrics); err != nil {
```
hawx, line 105: I think it would make sense to push this into the runner

internal/scheduled/runner.go
```diff
@@ -70,26 +76,67 @@ func NewRunner(logger Logger, store ScheduledStore, donorStore DonorStore, notif
...
 		if errors.Is(err, dynamo.NotFoundError{}) {
 			r.logger.InfoContext(ctx, "no scheduled tasks to process")
-			return nil
+
+			var metrics cloudwatch.PutMetricDataInput
+
+			if processedTasks > 0 || errorTasks > 0 || ignoredTasks > 0 {
+				elapsed := r.since(start)
+
+				metrics = cloudwatch.PutMetricDataInput{
```
hawx, line 98: it might be nicer to abstract this a little bit, you could have `(*Runner).Processed/Ignored/Errored()` methods to increment the counters and a `(*Runner).Metrics()` to return the `cloudwatch.PutMetricDataInput`, or something like that

cmd/scheduled-task-adder/main.go
```diff
@@ -0,0 +1,168 @@
...
+		return fmt.Errorf("failed to create dynamo client: %w", err)
+	}
+
+	const batchSize = 100
+	donorChan := make(chan any, batchSize)
+	eventChan := make(chan any, batchSize)
+
+	start := time.Now()
+
+	go func() {
```
hawx, line 72: I don't see why you've got this in a go routine with channels? But also what is the plan for this long term being a lambda?

acsauk: It was originally to try to get the time taken to add 1000s of tasks but I've just done a quick comparison with the existing version on main and the run times are within a second of each other 🤦  I'll revert to the original implementation.

I'm planning on having a CI task run to add a large amount of scheduled tasks and then hit a deployed version of the runner lambda so we can keep an eye on performance over time as we add more actions to it. It doesn't necessarily have to run in a lambda but it felt easier than applying CI permissions to run the adder against real AWS.

hawx (approved):

PR #1631, opened 2024-11-19
Fix template name for Check you can sign page

hawx:
Just spotted that I messed up this template name, and added a test so we'd spot problems in the future


acsauk (approved):
:shipit:

PR #1636, opened 2024-11-20
MLPAB-2686 Add method to ensure decisions are correct given current state

hawx:
since it needs to know about the state of everything, I've pushed the logic up to `donordata.Provided`


acsauk (approved):
:shipit:

PR #1638, opened 2024-11-21
MLPAB-2548 Prevent certificate provider going to pages they shouldn't

hawx:
I thought it would be better to split this PR out


acsauk (approved):
:shipit:

PR #1640, opened 2024-11-21
MLPAB-2321: Ignore scheduled tasks if LPA status is cannot-register

acsauk:
# Purpose

Fixes MLPAB-2321


acsauk:
Some things to think about with scheduled tasks going forward:

* Ignored tasks won't get rerun with the current implementation. How do we want to handle this going forward?
* Will we have a retention policy on handled tasks and/or tasks at a certain age? If not, it makes sense to give tasks linked to an LPA with cannot-register a handled (or something) status

internal/dynamo/client.go
```diff
@@ -87,6 +86,31 @@ func (c *Client) OneByUID(ctx context.Context, uid string, v interface{}) error
...
+	response, err := c.svc.Query(ctx, &dynamodb.QueryInput{
+		TableName: aws.String(c.table),
+		IndexName: aws.String(lpaUIDIndex),
+		ExpressionAttributeNames: map[string]string{
+			"#LpaUID": "LpaUID",
+			"#SK":     "SK",
+		},
+		ExpressionAttributeValues: map[string]types.AttributeValue{
+			":LpaUID": &types.AttributeValueMemberS{Value: uid},
+			":SK":     &types.AttributeValueMemberS{Value: scheduledPrefix},
```
hawx, line 99: would probably be worth appending a `#` on this

internal/dynamo/client.go
```diff
@@ -375,6 +399,32 @@ func (c *Client) DeleteOne(ctx context.Context, pk PK, sk SK) error {
 	return err
 }
 
+func (c *Client) DeleteManyByUID(ctx context.Context, keys []Keys, uid string) error {
```
hawx, line 402: I don't think there is any point having `uid` again, we should be able to trust the previous query

hawx: so could use `DeleteKeys`

internal/dynamo/client.go
hawx: would it be worth making this less specific as `AllByUIDAndPartialSK` or something?

acsauk: Yeah fair point 👍

internal/dynamo/client.go
acsauk: I know the likelihood of getting exactly the same timestamp + action type is low but as the keys aren't unique I wanted a failsafe there to be sure we weren't deleting unintended events. If you're comfortable with that risk feel free to drop this while I'm out and just rely on the keys 👍

hawx: every (PK, SK) is unique, so it won't be an issue. I have spotted `Put` doesn't use `Create`, so it would currently overwrite an existing scheduled event, I'm going to change that

PR #1683, opened 2024-12-13
MLPAB-2018 Add scheduled events to remind certificate provider

hawx:
Because when the scheduled event should occur can change I'm going for the approach of setting multiple scheduled events, then checking if it is valid to run or should be ignored. This is safe as expiry can only go up (from 6 months to 24), so in that case we'd be ignoring the event at +3 months.


acsauk (approved):
:shipit:

PR #1684, opened 2024-12-13
MLPAB-2018 Add a UID to correspondent and send with event

hawx:
To allow the event to work https://github.com/ministryofjustice/opg-event-store/pull/111


acsauk (approved):
:shipit:

PR #1688, opened 2024-12-17
MLPAB-2385: Only show certificate provider phone when it exists

acsauk:
# Purpose

Fixes MLPAB-2385


hawx (approved):
thought it might also be nice to have a way to get this with fixtures, and have a cypress test for it

hawx (approved):

PR #1690, opened 2024-12-18
MLPAB-2019 Send reminders to attorneys when they haven't signed

hawx:


internal/notify/to.go
```diff
@@ -34,6 +34,16 @@ func (t to) toEmail() (string, localize.Lang)  { return t.email, t.lang }
 func (t to) toMobile() (string, localize.Lang) { return t.mobile, t.lang }
 func (t to) ignore() bool                      { return t.ignored }
 
+// ToDonorOnly is only needed when we won't want the email to go to the
+// correspondent, normally we will use ToDonor.
+func ToDonorOnly(donor *donordata.Provided) To {
```
hawx, line 39: unrelated change, but I spotted I'd missed this when chatting to Laura about the use of `((Greeting))` in templates

acsauk (approved):
:shipit:

PR #1694, opened 2024-12-30
MLPAB-2692: Allow changing details after a failed vouch

acsauk:
# Purpose

Fixes MLPAB-2692


web/template/donor/your_details.gohtml
```diff
@@ -42,10 +43,14 @@
                         .Donor.Donor.FullName .Donor.CanChange true) }}
                 </dl>
 
-                {{ if not $hasNoIdentity }}
+                {{ if .Donor.IdentityUserData.Status.IsConfirmed }}
                     <p class="govuk-inset-text">{{ tr .App "someOfTheseDetailsCanNoLongerBeChanged" }}</p>
                 {{ end }}
 
+                {{ if $vouchInProgress }}
```
acsauk, line 50: I've added https://opgtransform.atlassian.net/browse/MLPAB-2778 to get some updated content.

internal/donor/donordata/provided.go
```diff
@@ -175,9 +175,14 @@ type Provided struct {
 	HasSentApplicationUpdatedEvent bool `hash:"-"`
 }
 
-// CanChange returns true if the donor can make changes to their LPA.
-func (p *Provided) CanChange() bool {
-	return p.SignedAt.IsZero()
+// CanChangePersonalDetails returns true if the donor can make changes to their FirstNames, LastName or DateOfBirth.
```
hawx, line 178: I'd prefer us to keep `CanChange` too then, and have `CanChangePersonalDetails` use that instead of `!p.SignedAt.IsZero`. I just don't trust that the condition won't change in future, and it is a bit nicer to read

acsauk: Fair point - I was on the fence as it was only checking one prop but, like you said, this could change.

hawx (approved):

PR #1708, opened 2025-01-08
MLPAB-2728: Add progress tracker voucher contact banner

acsauk:
# Purpose

Fixes MLPAB-2728


internal/donor/donorpage/progress.go
```diff
@@ -69,6 +69,26 @@ func Progress(tmpl template.Template, lpaStoreResolvingService LpaStoreResolving
...
+		if !donor.Tasks.ConfirmYourIdentity.IsCompleted() && donor.Voucher.FirstNames != "" {
+			var notification progressNotification
+
+			if donor.VoucherInvitedAt.IsZero() && !donor.Tasks.PayForLpa.IsCompleted() {
+				notification.Heading = "youMustPayForYourLPA"
+				notification.Body = appData.Localizer.Format(
+					"returnToTaskListToPayForLPAWeWillThenContactVoucher",
+					map[string]any{"VoucherFullName": donor.Voucher.FullName()},
+				)
+			} else {
```
hawx, line 81:
```suggestion
			} else if !donor.VoucherInvitedAt.IsZero() {
```

otherwise it will show before they are contacted

acsauk: Ah, good catch - I'd originally thought it wasn't possible to get to the progress tracker in this state but they could enter details, not send and then proceed to sign.

hawx (approved):

PR #1714, opened 2025-01-13
MLPAB-2642 Send vouched identity to lpa-store if lpa is submitted

hawx:
I've removed `donordata.Provided.SubmittedAt` as it could have been set when the Lpa wasn't sent to the store, instead I'm setting `lpadata.Lpa.Submitted` if the Lpa came from the store so it can be used to consistently check


internal/lpastore/lpa.go
```diff
@@ -271,6 +271,7 @@ type lpaResponse struct {
 
 func lpaResponseToLpa(l lpaResponse) *lpadata.Lpa {
 	data := &lpadata.Lpa{
+		InStore:      true,
```
hawx, line 274: I wanted to avoid using an inference on other fields and just give us a simple boolean to use

acsauk (approved):
:shipit:

PR #1716, opened 2025-01-13
MLPAB-2774 - Replace rules when they change

andrewpearce-digital:
# Purpose

Generate a hash for the rules file and use that as a suffix to the rule group name

Fixes MLPAB-2774

## Approach

- Fix count (was inverted)
- upgrade network module to use latest version

## Learning


acsauk (approved):
:shipit:

PR #1718, opened 2025-01-14
MLPAB-2730: Show banner on successful vouch

acsauk:
# Purpose

Fixes MLPAB-2730


internal/task/notification.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+//go:generate enumerator -type NotificationSubject -empty -trimprefix
+type NotificationSubject uint8
+
+const (
+	NotificationSuccessfulVouch NotificationSubject = iota
+)
+
+func (n NotificationSubjects) MarshalDynamoDBAttributeValue() (types.AttributeValue, error) {
```
acsauk, line 21: Dynamo was struggling to handle the marshalling of this as I kept getting `reflect.Copy: uint8 != task.NotificationSubject`. Hopefully, this is the most straightforward way of handling but give me a shout if I've missed something obvious.

hawx: There are 4 banners of this type to build, so I feel like I'd still be happy having bools for each. 

My initial idea was to have notifications pushed in, then deleted after seen. Kind of treat it like a cookie flash message type thing, but with `(LPA#, DONORNOTIFICATION#)` records. But with only 4 I think that would have been overkill too.

hawx:
image

it should be the green success banner style for this

hawx (approved):

PR #1720, opened 2025-01-14
MLPAB-2575 - use tls rules to make an allow list

andrewpearce-digital:
# Purpose

Only allow network egress to services we use

Fixes MLPAB-2575

## Approach

- add tls rules
- set drop rules to alert mode
- upgrade network module to add log insight query

## Learning

- https://docs.aws.amazon.com/network-firewall/latest/developerguide/rule-group-stateful-creating.html
- https://docs.suricata.io/en/latest/rules/http-keywords.html
- https://docs.suricata.io/en/latest/rules/tls-keywords.html
- https://docs.aws.amazon.com/network-firewall/latest/developerguide/suricata-examples.html#suricata-example-domain-filtering


hawx (approved):

acsauk (approved):
:shipit:

PR #1721, opened 2025-01-15
MLPAB-2793 - use pull through cache for otel collector

andrewpearce-digital:
# Purpose

Remove use of public ecr images, starting with otel collector (ssm agent is not being used at present).

Fixes MLPAB-2793

## Approach

- use pull through cache host for otel collector image


acsauk (approved):
:shipit:

PR #1727, opened 2025-01-16
MLPAB-2802 - drop traffic that is not in the allowlist

andrewpearce-digital:
# Purpose

Enable dropping of requests not in allow list

Fixes MLPAB-2802

## Approach

- set alarm rules to drop mode
- remove .amazon.com from allow list as all requests for aws should be going through a vpc endpoint


acsauk (approved):
:shipit:

PR #1728, opened 2025-01-16
MLPAB-2154: Show banner (and hide content) when LPA withdrawn

acsauk:
# Purpose

Fixes MLPAB-2154

### DIFF
--- a/internal/donor/donorpage/progress.go
+++ b/internal/donor/donorpage/progress.go
@@ -47,6 +47,18 @@ func Progress(tmpl template.Template, lpaStoreResolvingService LpaStoreResolving
 			Progress: progressTracker.Progress(lpa),
 		}
 
+		if !donor.WithdrawnAt.IsZero() {
+			data.InfoNotifications = append(data.InfoNotifications, progressNotification{
+				Heading: "lpaRevoked",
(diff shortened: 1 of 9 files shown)


internal/donor/donorpage/progress.go
```diff
@@ -47,6 +47,18 @@ func Progress(tmpl template.Template, lpaStoreResolvingService LpaStoreResolving
...
+		if !donor.WithdrawnAt.IsZero() {
+			data.InfoNotifications = append(data.InfoNotifications, progressNotification{
+				Heading: "lpaRevoked",
+				Body: appData.Localizer.Format(
+					"weContactedYouOnAboutLPARevokedOPGWillNotRegister",
+					map[string]any{"ContactedDate": appData.Localizer.FormatDate(donor.WithdrawnAt)},
+				),
+			})
+
+			return tmpl(w, data)
```
hawx, line 59: I think we should still show any other relevant notifications

hawx: but also I wanted to discuss on this ticket whether we should even do it, as the missing button seemed more "final"

hawx (approved):

PR #1730, opened 2025-01-17
MLPAB-2702 Store hash of sharecodes and control where plaintext is used

hawx:


internal/sharecode/sharecodedata/code.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"log/slog"
+	"net/url"
+
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/random"
+)
+
+func Generate() (PlainText, Hashed) {
```
hawx, line 12: this is the main change, replacing `random.String(12)` so you get a plaintext and hashed version that are then only used in the appropriate places

acsauk (approved):
:shipit:

PR #1759, opened 2025-02-05
MLPAB-2808 - remove dns firewall

andrewpearce-digital:
# Purpose

Remove DNS firewall as we are not using it

Fixes MLPAB-2808

## Approach

- remove module and module instance


acsauk (approved):
:shipit:

PR #1760, opened 2025-02-05
MLPAB-2643 - refactor opensearch into account level regional resource

andrewpearce-digital:
# Purpose

Make Opensearch serverless resources regional for failover

Fixes MLPAB-2643

## Approach

- Move opensearch file into region module
- create variables to pass resources and attributes in from root
- use moved blocks to relocate resources to their new address

## Learning

- https://developer.hashicorp.com/terraform/language/moved

### DIFF
--- a/terraform/account/region/outputs.tf
+++ b/terraform/account/region/outputs.tf
@@ -2,6 +2,6 @@ output "ecs_autoscaling_alarm_sns_topic" {
   value = aws_sns_topic.ecs_autoscaling_alarms
 }
 
-output "opensearch_lpas_collection_vpc_endpoint" {
-  value = aws_opensearchserverless_vpc_endpoint.lpas_collection_vpc_endpoint
-}
+# output "opensearch_lpas_collection_vpc_endpoint" {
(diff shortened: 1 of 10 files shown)


terraform/account/region/outputs.tf
```diff
@@ -2,6 +2,6 @@ output "ecs_autoscaling_alarm_sns_topic" {
   value = aws_sns_topic.ecs_autoscaling_alarms
 }
 
-output "opensearch_lpas_collection_vpc_endpoint" {
-  value = aws_opensearchserverless_vpc_endpoint.lpas_collection_vpc_endpoint
-}
+# output "opensearch_lpas_collection_vpc_endpoint" {
+#   value = aws_opensearchserverless_vpc_endpoint.lpas_collection_vpc_endpoint
+# }
```
acsauk, line 7: Do we need to keep this still?

acsauk (approved):
One potential comment block to drop (maybe?)

terraform/account/region/outputs.tf
andrewpearce-digital: thanks, no we don't. It's gone

acsauk (approved):
:shipit:

PR #1772, opened 2025-02-11
MLPAB-2271 Validate for no links in textareas

hawx:


acsauk (approved):
:shipit: 

This works but I'll be interested to see feedback from the wider team for scenarios not containing `://` and if that still invalidates an LPA.

hawx:
We don't have to worry about catching everything since any text will be checked over. But we do have to not block things that wouldn't invalidate it. So I think this is the fairest compromise.

PR #1774, opened 2025-02-12
MLPAB-2841 - create athena resources for alb access logs

andrewpearce-digital:
# Purpose

Add infra to support querying application load balancer access logs stored in s3

Fixes MLPAB-2841

## Approach

- Create s3 bucket for Athena results
- create Athena table, work group and query
- ~TODO: allow s3 decryption of alb access logs bucket~ access logs cannot use CMKs so encryption is fine as is
- ~TODO: use customer managed key for Athena results~ Ticketed

## Learning

- https://github.com/ministryofjustice/opg-sirius-infrastructure/blob/main/account/modules/alb_logging_bucket
- https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html
- https://docs.aws.amazon.com/athena/latest/ug/application-load-balancer-logs.html


terraform/account/region/load_balancer_logs_create_table.tpl
```diff
@@ -0,0 +1,57 @@
+CREATE EXTERNAL TABLE IF NOT EXISTS {PICK YOUR ENVIRONMENT NAME} (
```
acsauk, line 1: Do we need to choose the env name or is this fine?

andrewpearce-digital: we will when we use the query, but this is an account level resource and just a starter for ten. I may end up creating athena queries in the env level terraform per environment when I've written them as part of the investigation for Abbi.

acsauk (approved):
:shipit:

PR #1777, opened 2025-02-13
rename athena db to avoid broken db in production

andrewpearce-digital:
# Purpose

An error during the production account creation of the athena db has resulted in a db that cannot be updated or deleted. I've opted to create another alongside.

## Approach

- rename db

I've already checked and applied this to production, and checked preproduction is ok.

## Learning


acsauk (approved):
:shipit:

PR #1788, opened 2025-02-17
MLPAB-2830:  Add sirius event emitting to event-logger

acsauk:
# Purpose

Fixes MLPAB-2830

### DIFF
--- a/cmd/event-logger/event_sender.go
+++ b/cmd/event-logger/event_sender.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/sqs"
+)
+
+type sqsClient struct {
+	svc      *sqs.Client
+	queueURL string
+}
+
+type CloudWatchEvent struct {
+	Version    string          `json:"version"`
+	ID         string          `json:"id"`
+	DetailType string          `json:"detail-type"`
+	Source     string          `json:"source"`
+	Account    string          `json:"account"`
+	Time       time.Time       `json:"time"`
+	Region     string          `json:"region"`
+	Resources  []string        `json:"resources"`
+	Detail     json.RawMessage `json:"detail"`
+}
+
+func (c sqsClient) SendMessage(ctx context.Context, detailType string, detail json.RawMessage) error {
+	v, err := json.Marshal(CloudWatchEvent{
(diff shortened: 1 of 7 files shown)


cmd/event-logger/event_sender.go
```diff
@@ -0,0 +1,80 @@
+package main
```
hawx, line 1: you could add this to the `.codecov.yml` so it doesn't count

cmd/event-logger/event_sender.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+type ChangeConfirmedEvent struct {
+	UID       string `json:"uid"`
+	ActorType string `json:"actorType"`
+	ActorUID  string `json:"actorUID"`
+}
+
+func (c sqsClient) SendImmaterialChangeConfirmed(ctx context.Context, lpaUID, actorType, actorUID string) error {
+	detail, _ := json.Marshal(ChangeConfirmedEvent{
```
hawx, line 37: instead of redefining the event shapes, could we either have `?detailType=whatever&detail={json}` or `?detailType=whatever&detail.uid=x&detail.actorType=y` as the interface (I was thinking you could assume anything that isn't `detailType` is nested in `detail` but `uid-requested` nests again)

cmd/event-logger/main.go
```diff
@@ -167,5 +170,38 @@ func main() {
 		fmt.Fprint(w, "</tbody></table></body>")
 	})
 
+	http.HandleFunc("/emit-sirius-event", func(w http.ResponseWriter, r *http.Request) {
```
hawx, line 173: this could be `/emit` if we make it for all the events

cmd/event-logger/event_sender.go
hawx: and another option is to make this a `POST` and send a JSON body

cmd/event-logger/main.go
```diff
@@ -167,5 +170,38 @@ func main() {
...
+			detailType := r.PathValue("detailType")
+
+			var detail json.RawMessage
+			if err := json.NewDecoder(r.Body).Decode(&detail); err != nil {
+				log.Printf("failed to unmarshal %s: %v", detailType, err)
+			}
+
+			var sendErr error
+
+			switch detailType {
```
hawx, line 189: I'd remove this so it always sends the message, I basically don't want to ever have to edit this file. We should treat `event-logger` as something that could be extracted from this repo and be useful for anyone

hawx (approved):

PR #1792, opened 2025-02-18
MLPAB-2711: Email donor when paper certificate provider fails ID check

acsauk:
# Purpose

Fixes MLPAB-2711


cmd/event-received/sirius_event_handler_test.go
```diff
@@ -1926,3 +1926,155 @@ func TestHandleMaterialChangeConfirmed(t *testing.T) {
...
+	notifyClient := newMockNotifyClient(t)
+	notifyClient.EXPECT().
+		SendActorEmail(ctx, notify.ToDonor(provided), "M-1111-2222-3333", notify.InformDonorPaperCertificateProviderIdentityCheckFailed{
+			CertificateProviderFullName: "a b",
+			LpaType:                     "property and affairs",
+			DonorStartPageURL:           "app:///start",
+		}).
+		Return(nil)
+
+	bundle, err := localize.NewBundle("../../lang/en.json", "../../lang/cy.json")
```
acsauk, line 1964: This was the only way I could think of testing this without being able to make a mock for Localizer (I tried various combos of interfaces but couldn't get a valid implementation of the type with our current setup)

hawx: it is a bit of a pain, maybe we should make `localizer.Localizer` an interface? Then you'd easily be able to mock it

cmd/event-received/sirius_event_handler.go
```diff
@@ -432,3 +445,27 @@ func handleChangeConfirmed(ctx context.Context, client dynamodbClient, certifica
...
+	if err := json.Unmarshal(event.Detail, &v); err != nil {
+		return fmt.Errorf("failed to unmarshal detail: %w", err)
+	}
+
+	provided, err := getDonorByLpaUID(ctx, dynamoClient, v.UID)
+	if err != nil {
+		return fmt.Errorf("failed to get donor: %w", err)
+	}
+
+	if provided.Tasks.SignTheLpa.IsCompleted() {
```
hawx, line 460: I don't think we need this check, as we only send paper forms once the donor signs

hawx:
> If you do not take any other action, we will assume that you want to apply to the Court of Protection

interesting line in the email, but no mention in the ticket of us having to do anything for this... 🙃

cmd/event-received/sirius_event_handler.go
acsauk: Good point, my head is always in online world.

hawx (approved):

PR #1798, opened 2025-02-21
MLPAB-2804: Set sign task to in progress on language selection

acsauk:
# Purpose

Fixes MLPAB-2804


internal/donor/donorpage/task_list.go
```diff
@@ -226,6 +220,10 @@ func taskListSignSection(provided *donordata.Provided) taskListSection {
 		}
 	}
 
+	if !provided.WitnessedByCertificateProviderAt.IsZero() {
```
hawx, line 223: you don't really need this, since the other page would give `CanGoTo == false`

hawx (approved):

PR #1825, opened 2025-03-05
MLPAB-2615: Use KMS key alias when putting to S3

acsauk:
# Purpose

Fixes MLPAB-2615

### DIFF
--- a/terraform/environment/region/modules/app/ecs.tf
+++ b/terraform/environment/region/modules/app/ecs.tf
@@ -499,6 +499,10 @@ locals {
           // TODO: change this to be per env when we actually integrate
           value = "https://use-lasting-power-of-attorney.service.gov.uk/lpa/dashboard"
         },
+        {
+          name  = "S3_UPLOADS_KMS_KEY_ALIAS"
+          value = data.aws_kms_alias.reduced_fees_uploads_s3_encryption.name
+        },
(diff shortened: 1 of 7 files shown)


terraform/environment/region/modules/app/ecs.tf
```diff
@@ -499,6 +499,10 @@ locals {
           // TODO: change this to be per env when we actually integrate
           value = "https://use-lasting-power-of-attorney.service.gov.uk/lpa/dashboard"
         },
+        {
+          name  = "S3_UPLOADS_KMS_KEY_ALIAS"
+          value = data.aws_kms_alias.reduced_fees_uploads_s3_encryption.target_key_arn
```
andrewpearce-digital, line 504: this is providing the key arn (arn:aws:kms:eu-west-1:account-id:key/mrk-1111222233334444) rather than the alias (alias/custom-key) that points to the key, both are supported. 

However, the alias is not going to change and refers to the correct key within any region (and not cross-region). 

```suggestion
          value = data.aws_kms_alias.reduced_fees_uploads_s3_encryption.name
```

acsauk: Ah yeah my bad, I'd missed I was referencing the arn and not the name.

PR #1826, opened 2025-03-06
MLPAB-2382 Prevent check and send when identity shows donor is under 18

hawx:


internal/donor/donordata/provided.go
```diff
@@ -398,6 +398,17 @@ func (p *Provided) SigningDeadline() time.Time {
 	return p.SignedAt.AddDate(2, 0, 0)
 }
 
+// DonorSigningDeadline gives the date at which the LPA should be signed by the
+// donor once identity is confirmed, and whether it will be possible for them to
+// meet that deadline.
```
acsauk, line 403: The last part of the sentence doesn't seem to match the code. Were you previously checking if they would be 18 at some point in this func?

acsauk (commented):
One question on a comment still being applicable.

internal/donor/donordata/provided.go
hawx: indeed I was

acsauk (approved):
:shipit:

PR #1828, opened 2025-03-07
MLPAB-2381 Block check and send when donor gives date of birth under 18

hawx:


internal/donor/donorpage/your_date_of_birth.go
```diff
@@ -58,6 +58,14 @@ func YourDateOfBirth(tmpl template.Template, donorStore DonorStore) Handler {
 					return err
 				}
 
+				if provided.Donor.IsUnder18() {
+					from := r.FormValue("from")
+					r.Form = url.Values{}
```
hawx, line 63: this is a bit ugh, but needed a way to prevent `from` from happening

acsauk (approved):
:shipit:

PR #1879, opened 2025-04-03
MLPAB-2883 Ignore removed attorneys when checking if all signed

hawx:


cypress/support/commands.js
```diff
@@ -61,7 +61,7 @@ Cypress.Commands.add('visitLpa', (path, opts = {}) => {
 // Function to poll a page until element contains text or timeout occurs
 Cypress.Commands.add('waitForTextByReloading', (selector, expectedText) => {
     const options = {
-        timeout: 10000,
+        timeout: 20000,
```
acsauk, line 64: Oh dear

acsauk (approved):
:shipit:

cypress/support/commands.js
hawx: the timeout increases will continue until test flakiness improves 🤖

PR #1880, opened 2025-04-03
MLPAB-2945 - Use CMK for events emitted log group

andrewpearce-digital:
# Purpose

A CMK supports key policies to manage access to encrypt and decrypt

Fixes MLPAB-2945

## Approach

- use app logs cmk for emitted events log group

## Learning


acsauk (approved):
:shipit:

PR #1887, opened 2025-04-08
MLPAB-2400 Allow entering non-uk address for donor

hawx:


andrewpearce-digital (approved):

internal/donor/donordata/provided_test.go
```diff
@@ -607,6 +607,29 @@ func TestActorAddressesActorWithNoAddressIgnored(t *testing.T) {
 	assert.Equal(t, want, donor.ActorAddresses())
 }
 
+func TestActorAddressesActorWithNonUKAddressIgnored(t *testing.T) {
+	donor := &Provided{
+		Donor: Donor{FirstNames: "Donor", LastName: "Actor", Address: place.Address{
+			Line1:      "123 Rue Faux",
+			TownOrCity: "La Ville Fausse",
```
acsauk, line 614: lol

internal/onelogin/did_test.go
```diff
@@ -76,8 +76,20 @@ func TestGetDIDWhenRefreshAfterCacheControlDuration(t *testing.T) {
 
 	logger := newMockLogger(t)
 
-	client := getDID(ctx, logger, doer, "identity-url")
-	client.refreshRateLimit = time.Duration(0)
+	// create and start client to prevent data race on refreshRateLimit
```
acsauk, line 79: 🙌

internal/place/international_address.go
```diff
@@ -0,0 +1,50 @@
...
+		line2 = a.Town
+	}
+
+	return Address{
+		Line1: line1,
+		Line2: line2,
+		Line3: line3,
+		// putting region here is weird, but don't really have a better place to
+		// store in the current structure
+		TownOrCity: a.Region,
```
acsauk, line 46: Tricky one...as a gut feeling having town and region together feels marginally better than region displayed as town. Obvs this is edge of edge case and may not matter so lets see how this tests with some real addresses.

acsauk (approved):
:shipit:

internal/place/international_address.go
hawx: yeah we really need some real uses, I discussed with Greg and the fields aren't used for anything other than ordering the address. So (very) long term, maybe we'll move to a more generic "list of lines" format

PR #1908, opened 2025-04-23
MLPAB-2966 stop using deprecated state lock mechanism for terraform

andrewpearce-digital:
# Purpose

Using dynamodb for state locking is now deprecated.

Fixes MLPAB-2966

## Approach

- set `use_lockfile` to true in S3 backend setup for environment and account level terraform 

## Learning

- https://developer.hashicorp.com/terraform/language/backend/s3#state-locking


acsauk (approved):
:shipit:

PR #1911, opened 2025-04-23
MLPAB-2934 - Address invalid attribute combination in S3 lifecycle rule

andrewpearce-digital:
# Purpose

Lifecycle rules have some required attributes even if they are empty

Fixes MLPAB-2934

## Approach

- add a filter to the `abort multipart uploads` rule that uses the default prefix value for rules

## Learning

- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_lifecycle_configuration#filter


acsauk (approved):
:shipit:

PR #1917, opened 2025-04-25
MLPAB-2994 - Resolve invalid attribute combination warnings

andrewpearce-digital:
# Purpose

There are invalid attribute combination warnings for S3 lifecycle configurations

Fixes MLPAB-2994

## Approach

- add empty prefix filter for each s3 lifecycle configuration


acsauk (approved):
:shipit:

PR #1969, opened 2025-05-30
Align demo environment with other demo envs

gregtyler:
# Purpose

We now have a demo UID env and a demo Sirius env, so make MRLPA demo env talk to them:

FIxes SP-2848 #patch

## Approach

- Generate UIDs in the demo UID store
- Send events to the demo Sirius event bus
- Send reduced fees evidence to the demo bucket S3 bucket


gregtyler:
> I think we'll need to update the [firewall rules](https://github.com/ministryofjustice/opg-modernising-lpa/blob/main/terraform/account/region/network_firewall_rules.rules#L7) to have the new UID store address allow listed.

Good point, I've updated "development" -> "demo" and remove "integration" because there isn't an integration environment

andrewpearce-digital (dismissed):

andrewpearce-digital (approved):

PR #1970, opened 2025-05-30
MLPAB-2990: Replace inline name and date of birth errors with interruptions

acsauk:
# Purpose

Fixes MLPAB-2990


cypress/e2e/donor/add-correspondent.cy.js
```diff
@@ -84,4 +84,27 @@ describe('Add correspondent', () => {
                 });
             });
     });
+
+    it.only('warns when name shared with donor', () => {
```
hawx, line 88:
```suggestion
    it('warns when name shared with donor', () => {
```

hawx (dismissed):

cypress/e2e/donor/add-correspondent.cy.js
acsauk: Interesting, the stop-only pre-commit hook didn't pick this up 🤔

PR #2017, opened 2025-07-02
MLPAB-2173 Add some dynamo integration tests

hawx:
I wanted to prove that `Create` and `CreateOnly` were the same. Then thought I may as well extend to other things too, although this is not exhaustive yet, I left the methods that require indexes for another day and it is only happyish paths.


acsauk (approved):
:shipit:

PR #2041, opened 2025-07-16
reduce schedule to hourly

andrewpearce-digital:
# Purpose

Save costs while there are far fewer objects to replicate

## Approach

- set lamdab trigger schedule to 1 hourly

### DIFF
--- a/terraform/environment/region/modules/uploads_s3_bucket/lambda.tf
+++ b/terraform/environment/region/modules/uploads_s3_bucket/lambda.tf
@@ -59,8 +59,8 @@ data "aws_iam_policy_document" "s3_create_batch_replication_jobs" {
   provider = aws.region
 }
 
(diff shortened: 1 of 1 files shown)


terraform/environment/region/modules/uploads_s3_bucket/lambda.tf
```diff
@@ -66,7 +66,7 @@ resource "aws_scheduler_schedule" "invoke_lambda_every_15_minutes" {
     mode = "OFF"
   }
 
-  schedule_expression = "rate(20 minutes)"
+  schedule_expression = "rate(60 minutes)"
```
hawx, line 69: maybe the name(s) of this resource should also not mention a time?

andrewpearce-digital: good point, it's 3 or 4 changes old. fixed

PR #2074, opened 2025-08-07
MLPAB-3231: Use imposter-go for mock-pay

acsauk:
# Purpose

Fixes MLPAB-3231

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -183,6 +183,8 @@ services:
       context: ..
       dockerfile: docker/mock-pay/Dockerfile
     container_name: mock-pay
+    environment:
+      - IMPOSTER_LOG_LEVEL=TRACE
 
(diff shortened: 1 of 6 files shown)


docker/docker-compose.yml
```diff
@@ -183,6 +183,11 @@ services:
       context: ..
       dockerfile: docker/mock-pay/Dockerfile
     container_name: mock-pay
+    environment:
+      - IMPOSTER_LOG_LEVEL=TRACE
+
+    ports:
+      - "10001:8080"
```
hawx, line 190: do we need to keep the ports?

hawx (approved):

docker/docker-compose.yml
acsauk: No, this is from experimenting with the mock endpoints, I'll drop 👍

PR #2148, opened 2025-09-16
MLPAB-1500: Add privacy notice

acsauk:
# Purpose

Fixes MLPAB-1500


hawx (approved):
but feel free to merge if you disagree, as it probably won't be something we change that often

PR #2158, opened 2025-09-23
MLPAB-3305 Normalise capitalisation and hyphenation when matching names

hawx:
Also adds a similar change for address matching


acsauk (approved):
:shipit:

PR #2159, opened 2025-09-23
MLPAB-3296 Rewrite getDID tests to not rely on real time

hawx:


internal/onelogin/did.go
```diff
@@ -165,16 +165,13 @@ func (c *didClient) backgroundRefresh() {
 }
 
 func parseCacheControl(s string) time.Duration {
-	for _, directive := range strings.Split(s, ",") {
-		key, val, _ := strings.Cut(strings.TrimSpace(directive), "=")
-		switch key {
-		case "max-age":
+	for directive := range strings.SplitSeq(s, ",") {
```
hawx, line 168: linter suggested this change, then the rest is a pointless rewrite just because

PR #2168, opened 2025-09-30
MLPAB-3325 Set company number when signing in attorney fixtures

hawx:


acsauk (approved):
:shipit: 

My bad on this one. It feels like this has exposed a gap in some type of testing if I was able to merge the company number change without actually sending the data 🤔

andrewpearce-digital (approved):

PR #2182, opened 2025-10-06
MLPAB-3282 permissions

andrewpearce-digital:
# Purpose

Workspace Cleanup OIDC role could not assume terraform roles and have permission to cleanup resources not managed by terraform

Fixes MLPAB-3282

## Approach

- use new cleanup oidc role
- use vars for oidc role arns
- get shared token for pagerduty for local dev
- temp: don't delete workspaces so I can retry later
- temp: don't exit with error when trying to delete opensearch indices


acsauk (approved):
:shipit:

PR #2183, opened 2025-10-06
MLPAB-2681 Enforce uniqueness of keys when creating not reading

hawx:


internal/supporter/member_store.go
```diff
@@ -109,18 +119,21 @@ func (s *MemberStore) Create(ctx context.Context, firstNames, lastName string) (
...
 	}
 
-	if err := s.dynamoClient.Create(ctx, link); err != nil {
-		return nil, fmt.Errorf("error creating organisation link: %w", err)
+	transaction := dynamo.NewTransaction().
+		Create(member).
+		Create(link).
+		// In GetAny and OrganisationStore.Get we rely on there being only one
+		// member per session ID. This assumption may need to change in the future.
+		Create(dynamo.ReservedSK(member.SK))
```
acsauk, line 133: What was the scenario you were preempting here?

acsauk (approved):
:shipit:

internal/supporter/member_store.go
acsauk: Gotcha 👍

PR #2326, opened 2026-04-22
SP-3745 - disable send to metrics event rule #minor

AidanJ97:


docker/mainstream-content/Dockerfile
```diff
@@ -17,6 +17,9 @@ COPY --link docker/mainstream-content/alpine_image_hardening.sh /harden.sh
 
 COPY --from=fetched-repo app/register-lpa-prototype/package*.json ./
 
+# update uuid to 11.1.1 to fix security vulnerability via govuk-prototype-kit
+RUN npm pkg set overrides.uuid=11.1.1
```
hawx, line 21: in #2343 I ignored these vulnerabilities, so you could delete those from the .snyks

AidanJ97: would you not rather have this patched as it is in a production deployed image?

hawx (approved):

PR #2335, opened 2026-04-29
MLPAB-3694 - Refactor OIDC usage

andrewpearce-digital:
# Purpose

Change how oidc roles are referenced in PR and path to live workflows.

This PR won't deal with dispatch workflows yet.

Fixes MLPAB-3694

## Approach

- Use variables for oidc roles instead of secrets in top level workflows
- use inputs instead of secrets in workflow calls for oidc roles
- remove install aws cli workflow as it's already installed in the runner
- replace webfactory ssh agent with shared action
- update ingress management script to use oidc role
- pin ingress management script dependencies using uv


.github/workflows/dispatch_change_dns_target_region.yml
```diff
@@ -76,13 +75,14 @@ jobs:
...
+      - uses: ministryofjustice/opg-github-actions/actions/github-deploy-key@ceeda9d30ec7f6705c819d0b7664b9a79df99a3b # v4.8.0
         with:
-          ssh-private-key: ${{ secrets.OPG_MODERNISING_LPA_DEPLOY_KEY_PRIVATE_KEY }}
+          deploy_key: ${{ secrets.OPG_MODERNISING_LPA_DEPLOY_KEY_PRIVATE_KEY }}
       - name: Terraform Init
         run: terraform init -input=false
         working-directory: ./terraform/environment
-      - name: Terraform Apply DNS region target change for ${{ github.event.inputs.environment }}
+      - name: Terraform Apply DNS region target change for ${{
+          github.event.inputs.environment }}
```
acsauk, line 85: Should this be on one line?

.github/workflows/docker_job.yml
```diff
@@ -92,7 +92,8 @@ jobs:
           ref: ${{ inputs.checkout_tag }}
 
       - name: Warn missing linting action
-        run: echo "::warning title=MISSING-DOCKER-LINT::Please add the Docker lint workflow step when an allowed action is available."
+        run: echo "::warning title=MISSING-DOCKER-LINT::Please add the Docker lint
+          workflow step when an allowed action is available."
```
acsauk, line 96: And this? Although I'm guessing its fine if the action is working as expected 🤷

.github/workflows/schedule_update_opg_metrics_credentials.yml
```diff
@@ -2,8 +2,8 @@ name: "[Scheduled] Update OPG Metrics API key"
 
 on:
   workflow_dispatch:
-  schedule:
-    - cron: '10 2 * * *' # Every 2:10 a.m.
+    # schedule:
+    # - cron: '10 2 * * *' # Every 2:10 a.m.
```
acsauk, line 6: Does this need to be re-enabled (or removed)?

.github/workflows/dispatch_change_dns_target_region.yml
andrewpearce-digital: this is the github actions linter doing this, and I dislike it

.github/workflows/schedule_update_opg_metrics_credentials.yml
andrewpearce-digital: no, opg-metrics is being decommissioned and we need to stop sending metrics to it, so we might as well stop getting up to date keys for it.

.github/workflows/ui_test_job.yml
```diff
@@ -55,8 +58,10 @@ on:
       test_onelogin_password:
         description: 'Test OneLogin password'
         required: false
-      ingress_management_oidc_role_arn:
-        required: true
+
+permissions:
+  id-token: write
```
acsauk, line 63: Assuming this permission is already implicitly provided here by GitHub, so this is to make it explicit that its happening for future selves? Or is it a side-effect of the new OIDC flow?

andrewpearce-digital: I thought it was implicit, or inherited from the calling workflow, but they default to none if not specified. id-token is required for oidc

acsauk (approved):
:shipit:

PR #2338, opened 2026-04-30
MLPAB-3694 - fix path to live

andrewpearce-digital:
Fixes MLPAB-3694

## Approach

- fix var names for oidc roles on path to live


acsauk (approved):
:shipit:

PR #2347, opened 2026-05-13
Use new colour variant instead of old govuk-tint

hawx:
Missed this when upgrading the govuk-frontend version


terraform/account/opg_metrics.tf
```diff
@@ -1,48 +0,0 @@
-data "aws_default_tags" "current" {
-  provider = aws.eu_west_1
-}
-
-data "aws_secretsmanager_secret" "opg_metrics_api_key" {
-  name     = "opg-metrics-api-key/mrlpa-${data.aws_default_tags.current.tags.account-name}"
-  provider = aws.shared_eu_west_1
-}
-
-data "aws_secretsmanager_secret_version" "opg_metrics_api_key" {
```
hawx, line 10: Image

🤷 so I deleted this file

acsauk (approved):
:shipit:

PR #2370, opened 2026-05-21
MLPAB-3700 - replace deprecated hash and range for key_schema

andrewpearce-digital:
# Purpose

Multi-attribute keys for GSIs use the key_schema block instead of hash_key/range_key. The hash_key and range_key arguments are deprecated in favor of key_schema.

Fixes MLPAB-3700

## Approach

- replace deprecated hash and range keys with key_schema

## Learning

- https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/dynamodb_table#basic-example-containing-global-secondary-indexes-using-multi-attribute-keys-pattern


acsauk (approved):
:shipit:

PR #2384, opened 2026-06-26
MLPAB-3698 Introduce a new forms package and use in voucher/your-name

hawx:
This is intended to replace the existing form and validation packages.


internal/forms/form.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http"
+	"net/url"
+)
+
+type Parseable interface {
+	field() Field
+	Parse(url.Values)
+}
+
+func ParsePostForm(r *http.Request, fields ...Parseable) []Field {
```
acsauk, line 14: Does this need to be exposed?

internal/forms/validator.go
```diff
@@ -0,0 +1,19 @@
...
+//	type X { validators []validator[x] }
+//
+// instead of (the simpler):
+//
+//	type X { validators []func(x) Error }
+//
+// so that in the page tests we can use assert.Equals. Using bare funcs as the
+// validators means they cannot be compared with reflect.DeepEquals, and so the
+// tests will become much much more annoying. Even more annoying than having to
+// define a type per validator...
```
acsauk, line 16: 👍

web/template/layout/error-summary.gohtml
```diff
@@ -1,18 +1,35 @@
...
-          {{ range .Errors }}
-            <li>
-              <a href="#f-{{ .Name }}">{{ .Error.Format $.App.Localizer }}</a>
-            </li>
-          {{ end }}
-        </ul>
-      </div>
-    </div>
-  {{ end }}
+    {{ if hasErrors . }}
```
acsauk, line 2: OK, now `dot` makes sense 😅

acsauk (dismissed):
Nice, this feels better having the error scoped to the form itself.

internal/forms/form.go
hawx: good point, I should probably just roll it into `Form`'s method

acsauk (approved):
:shipit:

PR #2385, opened 2026-06-29
MLPAB-3698 Use new forms package for all voucher pages

hawx:


internal/forms/error.go
```diff
@@ -20,6 +21,12 @@ type Error interface {
 	Format(localizer Localizer) string
 }
 
+type ErrorMessage string
```
hawx, line 24: I'll see if I get inspiration on a better name for this when it is used more

internal/forms/yes_no_form.go
```diff
@@ -0,0 +1,28 @@
+package forms
+
+import (
+	"net/http"
+
+	"github.com/ministryofjustice/opg-modernising-lpa/internal/form"
+)
+
+var (
+	YesNoUnknown = form.YesNoUnknown
```
hawx, line 10: I'll move these types at the end of the refactoring

acsauk (approved):
:shipit:

